PR #395, opened 2014-05-14
support non-javascript-friendly identifiers

boehlke:
Alloy creates bad javascript code when id write something like <View id="my-view"/>. Problem is the '-'
The provided change fixes that problem.


jhaynie:
@skypanther any reason we can't merge this already?

skypanther:
Our internal discussion on this PR went something along these lines:

Alloy IDs are used to create variable names. Thus Alloy would attempt to turn an ID of `my-id` into something like `$.my-id`, which would get interpreted as a subtraction operation. We could work around this by implementing this PR. However, in general we don't typically implement workarounds for otherwise prohibited or antipattern techniques. For example:

* In Alloy, you can't use reserved words for ID and class names, just like you can't name variables with such words natively in JavaScript. We could probably make it work. But we don't because it would encourage bad techniques.
* In Titanium in general, you cannot use certain characters & words for app IDs due to requirements of Java package naming (JPN) rules. Specifically, you can't include a hyphen (or other punctuation) in your app name because that name is used to generate your default app ID. This is truly only a restriction on Android, because the other platforms don't require conformance with JPN rules. So, we could permit a hyphen or other characters in an app name, then warn you to fix the app ID (or just auto-generate an ID that is JPN-compliant). But we don't because there are other ways to use special characters in app names that work on all platforms.

Based on that thinking, Tony and I decided to let this PR sit. Yes, we could make `myid` work, but it encourages developers to further abstract their thinking from the underlying JavaScript code that Alloy generates. And, it would require developers to know to use `$[my-id]` to access such components, which increases documentation and learning requirements.

My apologies, we should have posted our rationale here and closed the PR rather than just letting appear to be ignored. If someone comes up with a strong, valid reason that this change is needed or that existing naming rules are insufficient, I will gladly reopen and accept this PR.

jhaynie:
@skypanther thanks for the update. makes sense to me.

PR #628, opened 2014-12-01
Adds ability to extend the Model and Collection from within the adapter

viezel:
This PR make the below possible:

```
module.exports.afterCollectionCreate = function(Collection) {
    // custom interface
    _.extend(Collection.prototype, {
        // extended functions and properties go here
        foo: function(){
            Ti.API.info("My custom adapter relevant functionality");
        }
    });

    return Collection;
};
```


viezel:
@ewanharris maybe a bit too late :D

PR #629, opened 2014-12-10
[ALOY-1198] Update ui/android_menu text for Honeycomb/action bar usage

tstatler:
- Updates comments to better reflect how Menu/MenuItem works by default with TI SDK 3.3.0 and later, as action bar items, not the options menu of yore).
- Conditionally displays the "Press the Menu button" label only if the API level is 10 or earlier (pre-Honeycomb).


skypanther:
**APPROVED**

PR #646, opened 2014-12-29
[ALOY-1206] Regenerate controllers when changing distr. target

skypanther:
https://jira.appcelerator.org/browse/ALOY-1206


feons:
**Tested and Approved.**

PR #647, opened 2015-01-05
[ALOY-1215] Alloy RSS Reader sample does not work on iPad

skypanther:
https://jira.appcelerator.org/browse/ALOY-1215


feons:
Looks like the failure is caused by missing ```reset:true``` flag in 
https://github.com/skypanther/alloy/blob/eaa898d0ea22c0800c20b3da665c04f2fef45cfb/test/specs/install.js#L24-L31

skypanther:
Yes, fixed by #649

feons:
**Approved.**

PR #648, opened 2015-01-06
regex change to support {{ }}

dbankier:
Alloy is currently catching and interpreting `{{ }}` as an alloy binding when it should only catch `{ }`.
This change is needed for https://github.com/dbankier/nano.

### DIFF
--- a/Alloy/commands/compile/styler.js
+++ b/Alloy/commands/compile/styler.js
@@ -16,7 +16,7 @@ var STYLE_ALLOY_TYPE = '__ALLOY_TYPE__';
 var STYLE_EXPR_PREFIX = exports.STYLE_EXPR_PREFIX = '__ALLOY_EXPR__--';
 var STYLE_REGEX = /^\s*([\#\.]{0,1})([^\[]+)(?:\[([^\]]+)\])*\s*$/;
 var EXPR_REGEX = new RegExp('^' + STYLE_EXPR_PREFIX + '(.+)');
-var BINDING_REGEX = /\{([^:}]+)\}/;
+var BINDING_REGEX = /(?:[^\{]|^)\{([^:}]+)\}(?:[^\}]|$)/;
 var VALUES = {
 	ID:     100000,
 	CLASS:   10000,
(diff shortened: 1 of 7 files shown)


FokkeZB:
The Travis CI build seems to have [failed](https://travis-ci.org/appcelerator/alloy/builds/46043907) over nothing, but @sfeather might be right that it also needs changes where the regexp is used because the full match now includes the character before and after. The compiler only [uses the regexp](https://github.com/dbankier/alloy/blob/double-brace/Alloy/commands/compile/styler.js#L520) on style values, so it might indeed work the way it is but it *does* do some replacements after line 520 that will now include the character before or after (if any, e.g. in values with multiple `{name}` in one value?)

dbankier:
went back to tony's original and used a negative lookahead instead.
So yes it will still catch `{{ name}` but `{{ name }}` is safe.

FokkeZB:
Good one ;)

skypanther:
Thanks @FokkeZB and @dbankier

The travis failure is unrelated. I thought it was simply a timeout issue, but it keeps popping up. I'll resolve that in [ALOY-1222](https://jira.appcelerator.org/browse/ALOY-1222)

skypanther:
I'd appreciate it if you'd add a test case to the PR. You could create a copy of the binding_no_persistence app and maybe add an extra label with non-binding {{foo}} text.

Having done that, I see that this PR works but not in all cases. For example, using the modified binding_no_persistence app:

```xml
<!-- adding before the table, this works as expected -->
<Label top="20" text="{{This is not a binding reference}}"/>

<!-- Then, trying some variations in the data-bound tableview -->
<TableViewRow title="{name}"/> <!-- works -->
<TableViewRow title="My hero is {name}"/> <!-- fails -->
<TableViewRow title="{name} and {{name}}"/>  <!-- fails -->
```

Log output includes:

```
[DEBUG] {title:typeof __alloyId2.__transform['name'] !== 'undefined' ? __alloyId2.__transform['name'] :
 My hero is '+__alloyId2.get('name') +',}
```

Looks like strings aren't being properly quoted.

Make sure to also create new _generated code before running the jake tests with:

```shell
node tools/create_generated_code.js testing/ALOY-1219
```

skypanther:
I meant to include a link to https://jira.appcelerator.org/browse/ALOY-443 where mixed binding string support was added.

dbankier:
@skypanther. Done and passing without a change.

These two are not valid test cases for this issue and a separate issue should be raised and the issue existed before:

~~~
<TableViewRow title="My hero is {name}"/> <!-- fails -->
<TableViewRow title="{name} and {{name}}"/>  <!-- fails -->
~~~

Simple reason is that alloy databinding never supported this:

~~~
<TableViewRow title="My hero is {name}"/> <!-- fails -->
~~~

jhaynie:
sweet. this will be cool

dbankier:
@jhaynie fast-track? :)

dbankier:
Just to update my comment before with a conversation with @skypanther:
screen shot 2015-01-08 at 9 20 52 am

dbankier:
Ok, I modified and squashed all the test commits. Done.

skypanther:
**APPROVED** and merged. Thanks @dbankier

dbankier:
Excellent! Thanks.

PR #659, opened 2015-01-29
[TIDOC-2058] Added @editurl JSDuck field that contains GitHub edit link for that class/API

tstatler:
To support https://jira.appcelerator.org/browse/TIDOC-2058.

Note: This step is done automatically for the Titanium docs, since they need to converted from YAML to JSDuck format anyway. Alloy docs are in JSDuck format already, so they are passed as-is to JSDuck for processing. As there are only a few Alloy doc sources, it made sense just to add them manually.

### DIFF
--- a/docs/apidoc/backbone.js
+++ b/docs/apidoc/backbone.js
@@ -1,6 +1,0 @@
-/**
- * @class Backbone.Events
- * @hide
- * Class from which all Alloy controllers are derived.
-
- */
--- a/docs/apidoc/backbone.js
+++ b/docs/apidoc/backbone.js
@@ -1,0 +1,6 @@
+/**
+ * @class Backbone.Events
+ * @hide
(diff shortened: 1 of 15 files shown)


docs/apidoc/backbone.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * @class Backbone.Events
```
bhatfield: This is showing up in the TOC.  Maybe use `@pseudo` to hide?

bhatfield:
@tstatler 
This looks fine except for Backbone.Event which is showing up in the TOC.

You will need Mr. Poulsen or somebody with merge permissions to commit this.

tstatler:
@bhatfield , thanks, I removed that file from the PR.

skypanther:
Passes all tests locally. 

**APPROVED**

PR #660, opened 2015-02-04
Revert "Added @editurl JSDuck field"

tstatler:
I’ve found a better way to make it work on the doctools side, so no need to muck with the Alloy source.

This reverts commit ec0542368652e8c92f9d746940aa43f13da38633.


skypanther:
**APPROVED**

PR #662, opened 2015-02-09
[ALOY-858] Theme i18n and platform

feons:
https://jira.appcelerator.org/browse/ALOY-858

### DIFF
--- a/Alloy/commands/compile/compilerUtils.js
+++ b/Alloy/commands/compile/compilerUtils.js
@@ -670,24 +670,32 @@ exports.mergeI18n = function(srcI18nDir, compileConfigDir, opts) {
 			}
 		} else {
 			if (fs.statSync(source).isFile()) {
-				var doc = U.XML.parseFromFile(outputPath),
-					root = doc.documentElement,
-					head = root.getElementsByTagName('string')[0],
-					sourcexml = U.XML.parseFromFile(source);
+				var outxml = U.XML.parseFromFile(outputPath),
+					root = outxml.documentElement,
+					sourcexml = U.XML.parseFromFile(source),
+					obj = {};
+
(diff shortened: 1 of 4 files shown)


skypanther:
@feons The PR isn't working for me. Is my procedure wrong?

1. Create a new Alloy app
2. Copy the contents of the alloy_repo/test/apps/advanced/themes folder to my app
3. Copy the alloy_repo/test/projects/HarnessTemplate/i18n folder to my app's project directory
4. Modify the app's project_name/i18n/en/strings.xml to define a single key `title` the value of which is `STOCK i18N`
5. Copy the i18n folder to each of the blue, green, and dark theme folders
6. Modify each of theme's i18n/en/strings.xml files to set the `title` key equal to the theme's name
7. Modify index.xml, changing the first label to `<Label id="title">L('title')</Label>`
8. Build the app

My expectation is that the title label in the app will output the currently-active theme (from the theme's i18n file). Instead, I get "STOCK i18N" 

Additionally, if I re-build the app and check the build/i18n folder, the strings.xml file contains multiple keys of the same name:

```xml
<?xml version="1.0" encoding="UTF-8"?><resources>
    <string name="title">STOCK i18N</string>
    <string name="title">GREEN THEME</string>
    <string name="title">BLUE THEME</string>
</resources>
```

We can't simply append keys. We either need to merge keys (replacing existing or adding new as necessary) or require developers to place an entire copy of the master i18n file in the theme and then simply copy it to the build folder in place of the stock i18n files.

feons:
@skypanther, updated hook to inject cli flags at the proper time, also fixed build/i18n multiple keys issue.
Thanks!

skypanther:
FR and CR pass

**APPROVED**!

PR #666, opened 2015-02-18
[ALOY-1238] Can't re-create index controller after renaming original one

feons:
https://jira.appcelerator.org/browse/ALOY-1238


skypanther:
**APPROVED**

PR #667, opened 2015-02-19
[ALOY-1249] Fix misleading message when controller not found

skypanther:
https://jira.appcelerator.org/browse/ALOY-1249


Alloy/lib/alloy.js
```diff
@@ -417,7 +417,18 @@ exports.createWidget = function(id, name, args) {
  * @return {Alloy.Controller} Alloy controller object.
  */
 exports.createController = function(name, args) {
-	return new (require('alloy/controllers/' + name))(args);
+	try {
+		return new (require('alloy/controllers/' + name.replace(/^\//, '')))(args);
+	} catch (err) {
+		if(OS_IOS && err.indexOf('architecture:') !== -1) {
```
pinnamur: Right now, we are adding a hack in Alloy code to address only iOS specific errors. If there is something wrong with Android platform sooner, we can't keep on adding multiple condition checks based on the different errors from the platform.

I'm not a fan of hacking Alloy code to show relevant error messages. We should figure out the changes that triggered this problem in SDK (or where ever) and try to address the root cause, rather than adding a hack in Alloy code.

Alloy/lib/alloy.js
```diff
@@ -417,7 +417,18 @@ exports.createWidget = function(id, name, args) {
  * @return {Alloy.Controller} Alloy controller object.
  */
 exports.createController = function(name, args) {
-	return new (require('alloy/controllers/' + name))(args);
+	try {
+		return new (require('alloy/controllers/' + name.replace(/^\//, '')))(args);
```
pinnamur: I like the idea replacing those additional slashes at the beginning of the controller file name - but do we have any examples about the failures while loading controllers ?

Alloy/lib/alloy.js
skypanther: See the Jira ticket. I note which change caused this error.

feons:
Thank you @skypanther for your pull request. 
You solution fixes the issue at run-time, which I think is valid. However I'm in favor of resolving the problem at compile time, it's been addressed in https://github.com/appcelerator/alloy/pull/712.

Please let me know your concerns.

skypanther:
This would prevent organizing controllers into subfolder. Closing in favor of other solutions.

PR #669, opened 2015-03-02
[ALOY-1250] Remove BB from Alloy

feons:
https://jira.appcelerator.org/browse/ALOY-1250


cheekiatng:
Looks good. CR and FT passed. Merging.

PR #671, opened 2015-03-21
[ALOY-1237] Added getEvent, getEvents, removeEvent and removeEvents methods in BaseController.js

k0sukey:
Added the get/remove event methods on defined in XML. But not modify $.destroy(). It looks like a $.destroy() for data binding.

### DIFF
--- a/Alloy/commands/compile/compilerUtils.js
+++ b/Alloy/commands/compile/compilerUtils.js
@@ -353,6 +353,7 @@ exports.generateNode = function(node, state, defaultId, isTopLevel, isModelOrCol
 
 		_.each(args.events, function(ev) {
 			var eventObj = {
+				id: args.id,
 				obj: isModelOrCollection ? state.args.symbol : args.symbol,
 				ev: ev.name,
 				cb: ev.value,
@@ -370,10 +371,15 @@ exports.generateNode = function(node, state, defaultId, isTopLevel, isModelOrCol
 			} else {
 				immediateTemplate = "<%= cb %>?" + theEvent + ":" + theDefer + "=true;";
 			}
+			var theListener = '';
+			if (eventFunc === 'addEventListener') {
+				theListener = _.template("$.__events.push({id:'<%= id %>',type:'<%= ev %>',handler:<%= cb %>});", eventObj);
+			}
 
 			// add the generated code to the view code and post-controller code respectively
 			code.content += _.template(immediateTemplate, eventObj);
 			postCode = _.template(deferTemplate, eventObj);
+			postCode +=　theListener;
 			exports.postCode += state.condition ? _.template(codeTemplate, {
 				condition: state.condition,
 				content: postCode
--- a/Alloy/commands/compile/compilerUtils.js
+++ b/Alloy/commands/compile/compilerUtils.js
@@ -362,7 +362,12 @@ exports.generateNode = function(node, state, defaultId, isTopLevel, isModelOrCol
 
 			// create templates for immediate and deferred event handler creation
 			var theDefer = _.template("__defers['<%= obj %>!<%= ev %>!<%= escapedCb %>']", eventObj);
-			var theEvent = (eventFunc === 'addEventListener') ? _.template("$.addListener(<%= obj %>,'<%= ev %>',<%= cb %>)", eventObj) : _.template("<%= obj %>.<%= func %>('<%= ev %>',<%= cb %>)", eventObj);
+			var theEvent;
+			if (eventFunc === 'addEventListener') {
+				theEvent = _.template("$.addListener(<%= obj %>,'<%= ev %>',<%= cb %>)", eventObj);
+			} else {
+				theEvent = _.template("<%= obj %>.<%= func %>('<%= ev %>',<%= cb %>)", eventObj);
+			}
 			var deferTemplate = theDefer + " && " + theEvent + ";";
 			var immediateTemplate;
 			if (/[\.\[]/.test(eventObj.cb)) {
--- a/Alloy/lib/alloy/controllers/BaseController.js
+++ b/Alloy/lib/alloy/controllers/BaseController.js
@@ -24,12 +24,6 @@ var Controller = function() {
 		} : self.__controllerPath;
 	}
 
-	function trackListener(obj, type, callback) {
-		var id = obj.id || (obj.id = Ti.Platform.createUUID());
-		self.__events.push({id:id,type:type,handler:callback});
-		return id;
-	}
-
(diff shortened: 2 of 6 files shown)


k0sukey:
I got this message in CLA signing...

```
{"success": false, "description": ". is a required parameter" "\" state \, "code": 400}
```

muhammaddadu:
Hey, the CLA is experiencing some downtime. 
You can alternatively sign the CLA [here](https://appcelerator.echosign.com/public/hostedForm?formid=25LBTXJ55452R).

k0sukey:
Thanks @muhammaddadu, i was signed.

FokkeZB:
@k0sukey any thoughts on how this could be extended to also allow the user to somehow also track (and thus make removable) events he adds in the controller? Like $.listenTo($.aView, "click", aListener) maybe?

jhaynie:
thanks @k0sukey !! :rocket:

k0sukey:
@FokkeZB hmm, maybe extend Backbone.js?

@jhaynie Your welcome :D

FokkeZB:
@k0sukey I've done a PR on your PR with some suggestions:
https://github.com/k0sukey/alloy/pull/1

k0sukey:
@FokkeZB Thanks,

> * Add support for manually adding tracked event listeners via $.addListener($.myView, 'click', doClick);.

OK, I see. Is it remove on later add the event? And if so, trackEvent method a good way.

> * Remove $.getEvent/getEvents() since they have not much value apart from testing.

Oops, I want to see the test code.
And __events as looks like private namespace. It's preferable a getter method. Want to check whether there is an event.

> * Combine $.removeEvent/removeEvents() into a single, more versatile $.removeListener().

I was the same as the $.getView(s) method :D

FokkeZB:
@k0sukey 

* Indeed, I changed that `$.addListener()` is now what the compiled code for `on<Event>` attributes also uses, but since it is now a public method, developers can use it in the controller as well, with the convenience of being able to remove them with `$.removeListener()` as well.

* I don't see much use for `$.getEvent()`. Why would a developer want to get the current events? In the test app I just use `__events` to validate the other 2 methods working. I know it's private, so a dev shouldn't use it, but for test it is fine.

* I know, but I think it's better to use the signature of BackBone, which is what I did. An instead of passing an ID to pass the object instead. Because `trackEvent` generates an id if it is not found, this will make it work for views without IDs, added by the dev through `$.addListener()` in the controller as well.

k0sukey:
@FokkeZB 

* trackListener was wrapped, because only called from $.addListener

* bug fixed $.addListener. It is an error to when you want to remove the added events in manually in the controller

* $.getListener() method. Manually when I add an anonymous callback function, there are times I want to get a handler.

k0sukey:
And does not dependency to Titanium API in Alloy, it was changed to _.uniqueId.

FokkeZB:
@k0sukey great changes, but the "fix" for `$.addListener()` I don't get. The method now not only adds the listener, but also adds the proxy (view) to `.__views`. This is beyond the method's signature and I see no reason to. The only thing we need for the proxy to have is an `id` property so we can link it back to the events in `$.removeListener()`. No need for it to be in `.$__views`.

k0sukey:
@FokkeZB Please look at the doOpen function of test code.

e.g. without this fix, you can not refer to the Ti.UI.Button object.
```js
var button = Ti.UI.createButton({ title: 'button' });
$.addListener(button, 'click', function(){
  $.removeListener(); // remove all
});
```

FokkeZB:
Why not? You can just pass "button" as the first arg. Addlistener sets a unique id property (if not found) to identify the proxy and __views isn't involved in this.

k0sukey:
If does not add proxy in __views.

1. Add button event

 ```js
$.addListener(button, 'click', function(){});
 ```

 1. Generate ID ```_.uniqueId('__trackId')``` for button
 2. Added ```__events```(id, type, handler)

2. remove all listeners

 ```js
// proxy, type, callback was omit
$.removeListener();
```

 1. __events in order
 2. ```__views[event.id]``` object undefined
 3. Application error

FokkeZB:
Ah. shoot, you're right. Sorry for not seeing that.

Mmm, maybe instead of having a (generated) id in `__events` we should store an actual reference to the object?

k0sukey:
I not good to certainly pollute the ```__views```. Changed so as to include the proxy to ```__events```.

FokkeZB:
Awesome, I've asked the engineers to have a look.

k0sukey:
Thanks, I've good discussion :+1:

feons:
Looks great! **APPROVED.**

falkolab:
Guys, it would be nice to update the documentation

FokkeZB:
@falkolab created TIDOC-2300

FokkeZB:
@k0sukey this has been a while but @falkolab seems to be right:
https://jira.appcelerator.org/browse/ALOY-1237?focusedCommentId=362271&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-362271

Shouldn't the template for {{$.destroy()}} always call {{$.removeListener()}} so the user only has to call {{$.destroy()}} when e.g. the window of the controller closes?

Or was this by design when you [said](https://github.com/appcelerator/alloy/pull/671#issue-63348289):

> Added the get/remove event methods on defined in XML. **But not modify $.destroy().** It looks like a $.destroy() for data binding.

falkolab:
Documentation clearly says:
> Frees binding resources associated with this controller and its UI components

http://docs.appcelerator.com/platform/latest/#!/api/Alloy.Controller-method-destroy

k0sukey:
@falkolab I think essentially it's Data Bindings. Or included in the target until the event handler? I think different. If the doing, in options? hmm, any good idea?

falkolab:
I dont know, I can call removeLestener on window close handler or cleanup controller method. But https://github.com/appcelerator/alloy/releases/tag/1.7.6  clearly describes the behavior that is not implemented. I think name of destroy method does not have direct relation to the Data Bindings. And because call removeListener() without any argument lead to unsubscribe all listeners, I see no reason not to add it to the destroy() method.

FokkeZB:
I lean towards that we should add `$.removeListener()` to the `$.destroy()` code. However, this will mean that any app will be effected, although it should be fine. We should at least clearly document it in the (general) release notes and guides.

/CC @feons your call!

FokkeZB:
Since Alloy 1.7.8 is already out, I've created a new ticket to discuss automatically calling `$.removeListener()` from `$.destroy()`:

https://jira.appcelerator.org/browse/ALOY-1303

PR #672, opened 2015-03-23
[CLI-461] properly handle windows node/alloy paths

tonylukasavage:
also handles the case of no global installation of alloy


feons:
FR & CR done. **Approved.**

PR #675, opened 2015-03-30
[ALOY-1259] Wrap code form `alloy.js` to protect global scope

FokkeZB:


feons:
This will break existing apps that rely on variables defined in alloy.js.

FokkeZB:
Yes it will, but it would enforce a best - or should I say excellent - practice to not pollute the global scope. There rarely is a reason to (can't think of any really) and it's one the things I ran into most frequently when doing consultancy for optimising and debugging apps.

I wish we'd had enforced this from day one and I'm still for enforcing it now and doing a good blog post on it when we do. We could even include a `config.json` flag to disable it, but it will only make sense if the default is to enforce and not the other way around.

ingo:
@FokkeZB is there a way we could figure out how many apps we are breaking with a change like this?

Related, I want to use semantic versioning with all projects, so this would likely be a bump to 2.0. If we do that, we should consider other breaking changes we would introduce at the same time.

Alloy/template/app.js
```diff
@@ -7,5 +7,10 @@ var Alloy = require('alloy'),
 	_ = Alloy._,
 	Backbone = Alloy.Backbone;
 
+(function(exports) {
```
FokkeZB: Not sure if this should be named `exports`. We've seen confusion in Alloy controllers where `exports` really is the controller's `this` and not the CommonJS exports people take it for.

Maybe `globals`

FokkeZB:
I don't think there is. I mean, of course you can parse the entire codebase tree and see if there are references to variables that are not defined locally but in {{alloy.js}}, but that would be too much overhead.

If we inform them in the release notes and do a blog post where we explain the best practice of not polluting scope as well as how to disable this safeguard through {{config.json}} we should be good.

ingo:
I'm not sure I want to make it configurable. What is a reason why we should allow that? It's just one more think for QE to test.

FokkeZB:
Because it's a breaking change. Default to the breaking change with the option of opting out leaves a way for people to hold off having to refactor their app (usage of global vars) until they have time to do so.

The QE test is simply checking that with the option the compiled {{app.js}} does not have the wrapping anonymous function right?

ingo:
Yes, it is just one check. But multiply that by many such checks...

My preference is that we use semantic versioning for what it is. If you don't want the breaking change, don't use 2.0 until you are ready. That said, I don't know how feasible that is now that Alloy is packaged with the CLI. I would prefer if you could specify your alloy compiler easily.

FokkeZB:
Yeah, I know Tony always wanted the option to have it selectable (or even installed locally) per project, but I guess that'll take some work ;)

ingo:
I'd prefer we spend the effort on that over adding configuration options. I like how Microsoft has finally learned that compatibility modes and other configuration options cause as many problems as they solve WRT holding on to bad practices: http://blogs.msdn.com/b/ie/archive/2014/11/11/living-on-the-edge-our-next-step-in-interoperability.aspx

FokkeZB:
I guess it's up to you ;) I'd be all for it!

FokkeZB:
Closing for inactivity

PR #680, opened 2015-04-06
[ALOY-1260] Windowsphone not in platforms list in manifest.json

feons:
https://jira.appcelerator.org/browse/ALOY-1260


cb1kenobi:
Looks good! **APPROVED**

PR #681, opened 2015-04-06
[ALOY-1260] Windowsphone not in platforms list in manifest.json

feons:
https://jira.appcelerator.org/browse/ALOY-1260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ The underscore.js library was upgraded from version 1.4.4 to version 1.6.0.
 
 ##### XML Markup Enhancements
 
-  * Support <ActionView/> as a child tag of the <MenuItem> object to set the `actionView` property
+  * Support &lt;ActionView/&gt; as a child tag of the &lt;MenuItem&gt; object to set the `actionView` property
     of `Titanium.Android.MenuItem`.  Only supports the Android platform.
 
 


cb1kenobi:
I'm not sure exactly how this solves the problem. You are just checking that assets directory exists before copying. This is not the problem.

The problem is you create the project with one version of the SDK that had a known platform (in the case of the ticket "windowsphone"), then changed your SDK to another that doesn't have that platform. Alloy just copies all platforms over and doesn't check if the SDK supports it.

The correct solution is for Alloy to determine the Titanium SDK being used to build the app (which isn't always the selected Titanium SDK), then open that SDK's manifest.json and read which platforms are supported, then only copy assets for those platforms.

skypanther:
If the platform name is supposed to be windowsphone, you need to update the files in https://github.com/appcelerator/alloy/tree/master/platforms to be named windowsphone instead of windows. There's a guide in the wiki (which I no longer have access to) that describes how to add platforms to Alloy.  There are some other constants, tiapp.xml templates, etc. that you'll have to also update. 

When I added Windows platform support, I was told the platform would be identified by "windows" for both windows phone and windows store. But, that did go back and forth a couple of times.

cb1kenobi:
@skypanther Not the issue. The issue here is Alloy is copying assets for platforms that is NOT supported by the Titanium SDK specified in the tiapp.xml. You can't do that. Alloy needs to determine what the target Titanium SDK is, determine what platforms it supports is, then only copy assets for those platforms.

feons:
Accidentally close this, created a new pr: https://github.com/appcelerator/alloy/pull/682

PR #687, opened 2015-05-13
[ALOY-1265] Update builtin moment.js library to 2.10.2 (or latest)

FokkeZB:
Use the version with all locales included so we don't have to patch all of them now they use require as well


kopiro:
Oh, great thanks!

FokkeZB:
@kopiro be aware that @feons has a better PR in development here:
https://github.com/appcelerator/alloy/compare/master...feons:ALOY-1265

Just need to add the language selection based on `Ti.Locale` there.

FokkeZB:
**DO NOT MERGE**

Better one in development by @feons:
https://github.com/appcelerator/alloy/compare/master...feons:ALOY-1265

kopiro:
I don't understand, why that PR should solve `require` problem?

FokkeZB:
It has all the locales patched. More work, but that way we don't have to include the full lib for all apps and also make the related changes to the compiler.

kopiro:
But, you have to copy manually the languages in `app/alloy/moment` directory?

FokkeZB:
No, Alloy does that for you.

kopiro:
@FokkeZB are you sure? Because in my `Resources` directory there's no alloy/moment dir.
Maybe it's another bug or my environment.

In fact, an iOS exception is raised due this issue.

FokkeZB:
When the Alloy compiled finds a `require('alloy/moment')` in your code, then it should copy the lib over, including any locales for languages you have in the `i18n` folder.

kopiro:
Yup, ok!
But you should definitely should write this in the doc.

:)

FokkeZB:
Like here?
http://docs.appcelerator.com/platform/latest/#!/api/Alloy.builtins.moment

kopiro:
Yes, but you should write that only "i18n/*" languages will be copied!
Sometimes (like now, in my app), you don't need to create it.

FokkeZB:
How do you mean? You want the app to have localisation for dates in a language that you don't provide localisation for when it comes to strings?

PR #689, opened 2015-05-26
[ALOY-1264] Support New iOS SplitWindow in XML Markup

FokkeZB:
https://jira.appcelerator.org/browse/ALOY-1264


FokkeZB:
@feons test ALOY-846 failed because https://github.com/appcelerator/alloy/blob/master/test/apps/testing/ALOY-846/views/index.xml resulted in a tab without a (split)window when ran on an iPhone. Fixed that.

PR #691, opened 2015-05-27
[ALOY-1272] Building an alloy project with Titanium CLI on Windows fails 1_6_X

feons:
https://jira.appcelerator.org/browse/ALOY-1272


pinnamur:
CR **passed**.

PR #692, opened 2015-05-27
[ALOY-1276] fixes deepExtend()

xavierlacot:
This PR relates to ALOY-1276. Basically, it fixes a bug introduced in 84c9470d9932a5ad762f8220db79802d91ec4c86 which makes Alloy's `deepExtend()` method useless.

It might be a good idea to simple revert commit 84c9470d9932a5ad762f8220db79802d91ec4c86, but it would involve digging into the bug ALOY-1153 in a closer way.


FokkeZB:
Looks good!

feons:
@xavierlacot, thank you for the PR! **APPROVED.**

PR #693, opened 2015-05-27
[ALOY-1277] made it possible to parse a TSS string directly, not only tss files

xavierlacot:
This PR fixes ALOY-1277

At the moment, [Alloy's styler](https://github.com/appcelerator/alloy/blob/master/Alloy/commands/compile/styler.js#L261) only allows to load styles from a file.

In several cases (alternative syntaxes transformations, for instance), it might be useful to be able to load styles from a string. This PR achieves such a goal in a rather simple way - it introduces a `loadStyleFromString()` method, which does the trick. The original `loadStyle()` method calls `loadStyleFromString()`.

### DIFF
--- a/Alloy/commands/compile/styler.js
+++ b/Alloy/commands/compile/styler.js
@@ -268,35 +268,40 @@ exports.loadStyle = function(tssFile) {
 			U.die('Failed to read style file "' + tssFile + '"', e);
 		}
 
-		// skip if the file is empty
-		if (/^\s*$/gi.test(contents)) {
-			return {};
-		}
-
-		// Add enclosing curly braces, if necessary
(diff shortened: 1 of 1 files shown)


feons:
@xavierlacot, thank you for the PR!
Could you please include a sample app to test the change?

xavierlacot:
Well, all of the existing sample apps already use Alloy's styler. I hardly see how to test this specific change in a sample app. Basically all I did was moving code in a new method to make it directly callable :-)

ingo:
It seems we should be able to repurpose whatever existing unit tests to allow us to test this too. I'd prefer a unit test over a sample app.

feons:
I do value the contribution, but there's no direct usage of the new method during compilation, style from string is not supported.

It would be great to see the feature, support style from string, being introduced together with this change, and also a test app to demonstrate the usage.

PR #694, opened 2015-06-01
Omit the / if a controller name starts with it

Topener:
If a controller name starts with `/` it doesn't work properly, or not at all. It also doesn't throw an error. This change fixes the issue if a controller starts with `/` as it would still make sense to support it.

In this case, both these cases will work the same. Currently the 2nd fails

```js
Alloy.createController('tasks/share');
Alloy.createController('/tasks/share');
```

### DIFF
--- a/Alloy/lib/alloy.js
+++ b/Alloy/lib/alloy.js
@@ -417,6 +417,7 @@ exports.createWidget = function(id, name, args) {
  * @return {Alloy.Controller} Alloy controller object.
  */
 exports.createController = function(name, args) {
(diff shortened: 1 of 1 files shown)


feons:
@Topener, thank you!
I'd appreciate it if you'd add a test case to the PR and also please file a JIRA for this.

Topener:
Hi, perfect. I'll just do that. 

I will however close this pull request and initiate a new one, so I can also adjust the docs & collections at the same time, as they suffer the same problem.

PR #695, opened 2015-06-02
The name of a controller, model or collection doesn't support names with / at start

Topener:
The name of the controller can also be a relative path. But to make it uniform (just like with images) you should be able to add a `/` in front of the name/path. It currently doesn't support that. This fix resolves that for controllers,models and collections, and adds to the documentation it can also be a relative path.

An example why this should be supported is because it is consistent for everything. Image names/paths with a leading `/` on Android is even required, and a controller name/path with a leading `/` is generating errors. This leads to unexpected issues, and confusion. 

This fix makes it consistent:

``` javascript
Alloy.createController('name'); // will work
Alloy.createController('/path/to/controller'); // will also work
$.imageView.image = '/path/to/image.png' // will (only) work
Alloy.createCollection('name'); // will work
Alloy.createCollection('/path/to/collection'); // will work
```

Consistent behaviour across all names/paths leads to expected functionality. Exceptions make it confusion. This PR fixes that

### DIFF
--- a/Alloy/lib/alloy.js
+++ b/Alloy/lib/alloy.js
@@ -412,11 +412,12 @@ exports.createWidget = function(id, name, args) {
  * @method createController
  * Factory method for instantiating a controller. Creates and returns an instance of the
  * named controller.
- * @param {String} name Name of controller to instantiate.
+ * @param {String} name Name of controller, or relative path to controller to instantiate.
  * @param {Object} [args] Arguments to pass to the controller.
  * @return {Alloy.Controller} Alloy controller object.
  */
 exports.createController = function(name, args) {
+	name = name.indexOf('/') === 0 ? name.substr(1) : name;
 	return new (require('alloy/controllers/' + name))(args);
 };
 
@@ -427,11 +428,12 @@ exports.createController = function(name, args) {
  *
  * See [Backbone.Model](http://docs.appcelerator.com/backbone/0.9.2/#Model) in the Backbone.js documentation for
  * information on the methods and properties provided by the Model object.
- * @param {String} name Name of model to instantiate.
+ * @param {String} name Name of model, or relative path of model to instantiate.
  * @param {Object} [args] Arguments to pass to the model.
  * @return {Backbone.Model} Backbone model object.
  */
 exports.createModel = function(name, args) {
+	name = name.indexOf('/') === 0 ? name.substr(1) : name;
 	return new (require('alloy/models/' + ucfirst(name)).Model)(args);
 };
 
@@ -443,11 +445,12 @@ exports.createModel = function(name, args) {
  * See [Backbone.Collection](http://docs.appcelerator.com/backbone/0.9.2/#Collection) in the Backbone.js
  * documentation for  information on the methods and  properties provided by the
  * Collection object.
- * @param {String} name Name of model to hold in this collection.
+ * @param {String} name Name of model, or relative path of model to hold in this collection.
  * @param {Object} [args] Arguments to pass to the collection.
  * @return {Backbone.Collection} Backbone collection object.
  */
 exports.createCollection = function(name, args) {
+	name = name.indexOf('/') === 0 ? name.substr(1) : name;
 	return new (require('alloy/models/' + ucfirst(name)).Collection)(args);
 };
 


feons:
@Topener, I am not able to reproduce the issues you reported. To facilitate the discussion, could you include some test cases to the PR?

Topener:
Hi @feons, it only works on iOS device

feons:
@Topener, the first parameter to ```Alloy.createModel()``` and ```Alloy.createCollection()``` is the ```name``` of the model file under ```app/models``` directory, not the path to the model file. I'd prefer a test app that proves otherwise.

Topener:
@feons if you make directories in the controller folder, you can just include `path/to/controller/in/a/folder.js`, and do a `Alloy.createController('path/to/controller/in/a/folder').`. This works, and has always worked. 

Also a neat way of organizing your app and splitting controllers up.

FokkeZB:
I think what @Topener means is that currently this does not work on an iOS device:

    Alloy.createController('/my/controller');

While this does work on Android and iOS Simulator.

So it's for the sake of parity and avoiding user bugs.

Topener:
Thanks @FokkeZB. That is exactly what I mean

feons:
@Topener, @FokkeZB, thank you for the clarification! :) 
However I don't think ```Alloy.createModel('path/to/model')``` and ```Alloy.createCollection('path/to/model')``` are valid in the first place. Thoughts?

Topener:
Well according to the docs, they are supposed to be names. But Paths have been working for always. I also think this needs to be explained as Valid (hence the adjustment to the docs) so people can organise their projects better. Because not everyone knows about it, but everyone should. Because big projects just won't work without folders.

FokkeZB:
@feons I agree with @Topener that though not documented, it has been a best practice to group controllers this way and I see no reason why models/collections wouldn't. I think we should even document en promote it as best practice in the docs.

skypanther:
I fully agree that Alloy should support the leading slash in the `Alloy.createController()` method. I had a (poorly implemented) PR to add that support (which I've closed because it would have broken other things). 

I'm not so sure about the models/collections though. If I were still on the Alloy team, I'd be requesting that you break this into two PRs: one for controllers and one for models/collections. I'd merge your controllers PR immediately. But, I'd want to gather more information from the Support folks, community, etc. to see whether changing the models/collections was worth doing. 

I'm not saying it's a bad idea. Consistency and flexibility are good. But, somehow the models/collections change just feels wrong. So, I'm saying I'd approach the change with caution.

FokkeZB:
I'm 100% for this PR. Didn't we all agree that Alloy 1.x introduced more conventions/layers then needed? We should put as little in the way between the user and the `require()` that Alloy in the end does.

I assume in Alloy 2.x we would do something like this anyway:

*controllers/index.js*
```
var collection = require('models/myModel');
``` 

*views/index.xml*
```
<Collection src="models/myModel" />
```

*models/myModel.js*
```
class MyModel extends AlloyModel {
  ..
}

module.exports = new MyModel;
```

PS: I have no idea what I'm doing ^^ since I don't use ES6 yes but you get the idea. Instead of wrapping stuff like we do now we would provide classes the user extends, allowing him to directly require the actual model commonjs module.

Topener:
Can the related issue be reopened please, it says PR was merged, it wasn't. Just noticed this: https://jira.appcelerator.org/browse/TC-5548

FokkeZB:
Done and moved to Alloy project

feons:
I absolutely understand the idea here and I’m completely agreeable to making good improvements. 

1. the initial PR was good, it handled only leading slash in the **Alloy.createController()**. I was about to merge it once a test was included into the PR. 

2. the subsequent update made additional changes to **Alloy.createModel()** & **Alloy.createCollection()**. I get the idea, but does it work? Can you please confirm with a **test app** that the call to *Alloy.createCollection('path/to/collection');* correctly creates the underlying data table?

Point 2, to be able to organize model files in directories requires far more changes than just the leading slash, one thing off the top of my head is to handle the database table name, and there could be more.

I would really appreciate it if we could separate *controller*, *model & collection* into different tickets and get more opinions from the community.

FokkeZB:
@foens I'm fine with splitting the requests/PRS but the only places I could find where the name of the model is used other then for requiring the right file are:

* beforeModelCreate https://github.com/appcelerator/alloy/blob/master/Alloy/lib/alloy.js#L154
* afterModelCreate https://github.com/appcelerator/alloy/blob/master/Alloy/lib/alloy.js#L168

And the only (built-in) sync adapter that uses that name is sql, where it is no problem because it's simply used as cache key:

https://github.com/appcelerator/alloy/blob/master/Alloy/lib/alloy/sync/sql.js#L472-L513

Apart form that, the name/path of the model is irrelevant. It's {{db_name}} and {{collection_name}} (which are set by the user and unrelated to the model name/path that determine storage locations.

foens:
@FokkeZB You have annotated the wrong person here. You should have addressed @feons instead of @foens. He should be notified now.

FokkeZB:
The @foens!

ewanharris:
I took a look into this, the `createController` calls are now successful regardless of a leading slash or not in the path, I imagine this is now fixed due to the improved require work Chris W did.

`createCollection` and `createModel` calls however do fail and I have filed https://jira.appcelerator.org/browse/ALOY-1649 to track this.

I think it's best to deal with those issues in a new PR so I'm going to close this PR but thanks for the work @Topener!

PR #699, opened 2015-06-18
[TIMOB-18550] Update minimum Node.js version to 0.10

ingo:


feons:
This has been addressed in [ALOY-1239] (https://jira.appcelerator.org/browse/ALOY-1239).

PR #702, opened 2015-06-28
Make input text for Add feature readable on Android

vuinguyen:
Add font color 'black' to #itemField. As of Ti SDK 4.0, you can't read what you are typing into the text field otherwise. Seems to be a problem with only Android.


feons:
@vuinguyen, thank you for the PR.

feons:
**Approved.**

PR #704, opened 2015-07-01
Fixed a style ordering bug

xavierlacot:
This bug is described in [[AC-189]](https://jira.appcelerator.org/browse/AC-189).

In some cases, when the developer overloads in a given stylesheet some styles defined in the `app.tss` global style file, the alloy compilation won't produce the same resulting javascript accross successive compilations.

Alloy's styler orders style rules to define their priority, but there's a bug with the `app.tss` global styles being cached and therefore not correctly being counted by the styler when they're loaded back from cache, during a latter compilation phase when `app.tss` didn't change.

This fix simply correctly increments the styleCounter with the number of rules loaded from the global style cache. Kudos to my colleague @vdesdoigts for spotting the bug and isolating a reproducible symptom.


vdesdoigts:
:+1:

FokkeZB:
Verified the PR to work, please merge.

PR #711, opened 2015-07-17
[ALOY-1231] Mapping sample: add readme, remove map key

feons:
https://jira.appcelerator.org/browse/ALOY-1231


FokkeZB:
We should remove this sample app when https://github.com/appcelerator-developer-relations/sample.mapping/tree/alloy is merged.

PR #712, opened 2015-07-27
[ALOY-1249] iOS: Misleading error message when controller file not found

feons:
https://jira.appcelerator.org/browse/ALOY-1249


FokkeZB:
@feons why was this closed? Do you plan on another way to resolve https://jira.appcelerator.org/browse/ALOY-1249 ?

skypanther:
Agreed. This is still an issue. Fixing at compile time is the better solution than the runtime fix I proposed.

feons:
@FokkeZB, it began with an internal monologue, I wasn't quite satisfy with my solution. Instead of working on a clean slate I should continue working on this one.

skypanther:
FYI, this is potentially a breaking change for some developers. Your change would prevent someone from using runtime techniques to define a controller.

```javascript
// if this throws a compile-time error
Alloy.createController('doesNotExist');

// then these would too, even though they're valid techniques

// Technique A
var isLoggedIn = true;
Alloy.createController(isLoggedIn ? 'controllerA' : 'controllerB');
/* of course, this could be easily rewritten as
if (isLoggedIn) {
  Alloy.createController('controllerA');
} else {
  Alloy.createController('controllerB');
}
*/

// Technique B
var val = 'exists'; 
Alloy.createController('file' + val); // where fileexists is a valid controller name
// depending on how many values `val` could take, you could end up with
// some pretty verbose code to use static strings for all the permutations
```

Rather than halting the compile with an error, you could log out a warning message. Though, log messages will get lost in the shuffle of other compile messages. But at least it won't prevent users from using these valid techniques.

Or, use your PR as-is but add a compiler flag that says to ignore missing controllers so that developers have to make a conscious choice if they use these dynamically generated controller declarations.

Or, make sure that this is documented as a breaking change in the next release

FokkeZB:
@skypanther yup, we can't do this on compile time because of the reasons you point out. It would also just make the Alloy compile slower. I'm for the runtime solution.

skypanther:
I'd sacrifice compile time performance for runtime though. Feon indicated by chat that the Technique B I show above is not a problem with her change. And since Technique A is easily coded around, I'd say this should be merged. But I'd still add a note in the Release Notes that this change might need some code changes if you've used these dynamic declaration techniques (admittedly, probably not many do this).

FokkeZB:
Technique B as you have it is not because uglily will evaluate that. But if you compose the variable in slightly more complex ways it will fail.

And if you think if it, without the architecture thing the runtime red screen isn't that bad at all. So I say we better fix https://jira.appcelerator.org/browse/TIMOB-19881

skypanther:
Right, my intent was that there'd be some runtime logic determining which controller is loaded. The simplistic example I showed does not demonstrate that; and Feon's quick check of her implementation may not have tested that case.

feons:
Yes I admit that my fix is inconsistent when it comes to the run-time techniques Tim illustrated. It throws an error at compile time only when the parameter is a string. 

If you file a pull request that handles this at run-time, i'm happy to merge it.

FokkeZB:
I'd rather see https://jira.appcelerator.org/browse/TIMOB-19881 resolved and not do this at all ;)

feons:
@FokkeZB, good point!

PR #713, opened 2015-08-18
[ALOY-1297] Cannot use localized strings for AlertDialog ButtonNames in XML

feons:
https://jira.appcelerator.org/browse/ALOY-1297


feons:
**Approved.**

PR #714, opened 2015-08-18
[ALOY-1299] Increased the resolution of the default appicon@2x so tha…

cb1kenobi:
…t all other missing icons will be generated.


cb1kenobi:
**APPROVED**

PR #715, opened 2015-08-20
[CLI-768] Alloy app fails to build on Windows with error "Alloy compiler failed"

feons:
https://jira.appcelerator.org/browse/CLI-768
Dependent: https://github.com/appcelerator/appc-cli-titanium/pull/126


cheekiatng:
APPROVED

PR #717, opened 2015-08-20
[ALOY-1300] iOS: Assets Catalog inconsistency between Alloy and Classic

feons:
https://jira.appcelerator.org/browse/ALOY-1300


cheekiatng:
APPROVED!

PR #718, opened 2015-08-27
[ALOY-1285] TSS conditional statement is ignored by items dynamically generated on runtime

feons:
https://jira.appcelerator.org/browse/ALOY-1285


feons:
Merged.

PR #719, opened 2015-09-02
[ALOY-1241] Widgets are initialized twice in TableViewSection

feons:
https://jira.appcelerator.org/browse/ALOY-1241


feons:
*Approved.*

PR #721, opened 2015-09-03
[ALOY-1210] Add support for DIST in the Project Configuration File

feons:
https://jira.appcelerator.org/browse/ALOY-1210

### DIFF
--- a/Alloy/commands/compile/compilerUtils.js
+++ b/Alloy/commands/compile/compilerUtils.js
@@ -839,7 +839,7 @@ function generateConfig(obj) {
 }
 
 exports.parseConfig = function(file, alloyConfig, o) {
-	var j;
+	var j, distType;
 	try {
 		j = jsonlint.parse(fs.readFileSync(file,'utf8'));
 	} catch (e) {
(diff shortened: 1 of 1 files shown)


skypanther:
Thank you @feons! Is there a way this can be supported for Android builds? Android itself and the `ti build -p android` command don't support the concept of ad-hoc builds. But, our use-case calls for beta builds signed using our company's certificate (rather than the developer's). To do that, we have to create store builds for Android (test builds use Appc's test certificate). It would be great to do "ad-hoc" builds for Android without resorting to swapping out configuration files.

FokkeZB:
@feons we could check if deployType = production and target <> dist-playstore I guess?

feons:
@FokkeZB, ```deployType = production && target = dist-playstore``` that would be a store build for Android isn't it? 

@skypanther, ```dist:store``` will apply to both iOS and Android, if you have different distribution configurations for each platform, you could do ``` "os:ios dist:store" ```, ```"os:android dist:store" ```.  
As you mentioned, ```ti build -p android``` doesn't have ```--target``` value ```dist-adhoc```, if you have to do an ""ad-hoc"" build for Android, you will have to use ```alloy compile --config platform=android,deploytype=production,target=dist-adhoc``` and include ```"os:android dist:adhoc"``` in ```config.json```

FokkeZB:
@feons I meant  `config.deployType === 'production' && config.target !== 'dist-playstore'`

Since Alloy compiles as part of Titanium I don't think you can change the config Titanium passes on to the Alloy CLI.

feons:
Merged.

PR #728, opened 2015-10-08
[ALOY-1310] Alloy: Picker not working in ListView template

feons:
https://jira.appcelerator.org/browse/ALOY-1310


pinnamur:
@feons you need to update version

pinnamur:
CR approved.

PR #730, opened 2015-10-08
[ALOY-1314] Added iOS specific default icon and updated the logic whe…

cb1kenobi:
…n creating new projects to copy all default icons into the project.


feons:
Looks good. **APPROVED**

PR #733, opened 2015-10-19
[ALOY-1319] Support CardView in XML

feons:
https://jira.appcelerator.org/browse/ALOY-1319

### DIFF
--- a/test/apps/testing/ALOY-1319/views/index.xml
+++ b/test/apps/testing/ALOY-1319/views/index.xml
@@ -1,0 +1,8 @@
+<Alloy>
+	<Window>
+		<CardView ns="Ti.UI.Android" platform="android">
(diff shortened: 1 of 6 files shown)


test/apps/testing/ALOY-1319/views/index.xml
```diff
@@ -0,0 +1,8 @@
+<Alloy>
+	<Window>
+		<CardView ns="Ti.UI.Android" platform="android">
```
FokkeZB: Why the ns-attribute in this test? Shouldn't it work without?

FokkeZB:
PR **APPROVED** with SDK 5.1.0.v20151019145227 but the sample app should be updated to remove the `ns` attribute.

PR #734, opened 2015-10-20
[ALOY-1311] Alloy.jmk compile:app.js event never fires

feons:
https://jira.appcelerator.org/browse/ALOY-1311


feons:
**Approved.**

PR #736, opened 2015-10-29
[ALOY-1320] Styles not applied when an element uses another namespace

feons:
https://jira.appcelerator.org/browse/ALOY-1320


FokkeZB:
Works like a clock!

feons:
@FokkeZB  thank you!
*Merged*

PR #742, opened 2015-11-19
[ALOY-1323] Alloy: module-tag in Alloy element effects all controller views

feons:
https://jira.appcelerator.org/browse/ALOY-1323


feons:
**Approved.**

PR #744, opened 2015-11-27
Use Alloy's deepExtend to handle conditional on-device style inheritance

xavierlacot:
Alloy selectors provide ways to overload style rules based on switches (formFactor, OS, conditional rules, etc.). Say the TSS is somehow:

```
...
".title-home": {
	bottom: 4,
},
".title-home[formFactor=tablet]": {
	bottom: 8
},
...
```

The generated code will look like:

```
Alloy.deepExtend(true, o, {
    bottom: 4
});
Alloy.isTablet && _.extend(o, {
    bottom: 8
});
```

If this works fine with most of the properties, this is particularly annoying with font or nested properties. For instance, 

```
".title-home": {
	font: {
		fontFamily: "Delius-Regular",
		fontSize: 11
	}
},
".title-home[formFactor=tablet]": {
	font: {
		fontSize: 22
	}
},
```

will result in the font name being simply ignored on tablets, as `_.extend()` is not a deepExtend method. We're lucky, Alloy provides a `deepExtend()` implementation which we may want to use. This is what this PR does, allowing a much more easier class mix:

    <Label class="fonticon icon-row-right icon-arrow-md icon-arrow-right white" />

All styles are merged deeply, and it becomes unnecessary to repeat over and over the font name.

I *know* that this PR may break many tests, but I strongly believe that this behavior is the right way styles *should* work, in the sake for development efficiency.

The JIRA ticket is [AC-515](https://jira.appcelerator.org/browse/AC-515)


feons:
@xavierlacot, thank you for the PR!
 Looks good! **Approved.**

PR #745, opened 2015-12-04
moment.lang() example fix

mcvendrell:
Currently, moment need a 2-letter language to setup it right, so cannot use moment.lang(Ti.Locale.currentLocale) (2 language- 2 country)
Instead, we must use moment.lang(Ti.Locale.currentLanguage) (2 language only)

### DIFF
--- a/docs/apidoc/moment.js
+++ b/docs/apidoc/moment.js
@@ -15,7 +15,7 @@
  *     var moment = require('alloy/moment');
  *     require('alloy/moment/lang/de');
  *     require('alloy/moment/lang/fr');
- *     moment.lang(Ti.Locale.currentLocale); // Set current system locale, as a combination of ISO 2-letter language and country codes.
+ *     moment.lang(Ti.Locale.currentLanguage); // Set current system locale, as a combination of ISO 2-letter language.
  * 
  * For documentation, usage examples and more information, see [http://momentjs.com/](http://momentjs.com).
  */


mcvendrell:
Wait to do this PR, it seems that locale dialects also has 4 letters, but in lowercase. Need more tests.

falkolab:
This is tested code. Please use it for doc update:
```javascript
var moment = require('alloy/moment');

// First add languages that support your application
var languages = ['ru', 'nl', 'fr', 'fr-ca'];
// Full language list:  https://github.com/appcelerator/alloy/tree/master/Alloy/builtins/moment/lang

// You must explicit way to require language files in order to compiler saw their. 
require('alloy/moment/lang/ru');
require('alloy/moment/lang/nl');
require('alloy/moment/lang/fr');
require('alloy/moment/lang/fr-ca');

function selectLanguage(lang) {
	// Not all locales with country code has corresponding file in the alloy/moment. 
	// So we will use first 2-letter language code as insurance.
	[lang.split('-')[0], lang.toLowerCase()].forEach(function(lang) {
		if(languages.indexOf(lang) !== -1) {			
			moment.lang(lang);
		}		
	});
}

selectLanguage('fr-ca');
```
Test code:
```javascript
Ti.API.info('current locale:', Ti.Locale.currentLocale);
Ti.API.info('current language:',  moment.lang(), 'date:', moment().format('L'));

selectLanguage(Ti.Locale.currentLocale);
Ti.API.info('current language:',  moment.lang(), 'date:', moment().format('L'));

// Ask user for language or select as you need (Ti.Locale.currentLocale).
// Letter case does not matter.
selectLanguage('fr-ca');
Ti.API.info('current language:',  moment.lang(), 'date:', moment().format('L'));

selectLanguage('fr');
Ti.API.info('current language:',  moment.lang(), 'date:', moment().format('L'));

selectLanguage('nl-BE');
Ti.API.info('current language:',  moment.lang(), 'date:', moment().format('L'));
```

FokkeZB:
I believe MomentJS should fallback from more specific to less specific for you:

> Moment will also try locale specifier substrings from most-specific to least-specific until it finds a locale it knows. This is useful when supplying Moment with a locale string pulled from the user's environment, such as window.navigator.language.

https://github.com/appcelerator/alloy/blob/master/Alloy/builtins/moment.js#L894-L914

So it looks like the existing `moment.lang(Ti.Locale.currentLocale);` is OK. Or does your test show otherwise?

FokkeZB:
I believe MomentJS should fallback from more specific to less specific for you:

> Moment will also try locale specifier substrings from most-specific to least-specific until it finds a locale it knows. This is useful when supplying Moment with a locale string pulled from the user's environment, such as window.navigator.language.

https://github.com/appcelerator/alloy/blob/master/Alloy/builtins/moment.js#L894-L914

So it looks like the existing `moment.lang(Ti.Locale.currentLocale);` is OK. Or does your test show otherwise?

FokkeZB:
I believe MomentJS should fallback from more specific to less specific for you:

> Moment will also try locale specifier substrings from most-specific to least-specific until it finds a locale it knows. This is useful when supplying Moment with a locale string pulled from the user's environment, such as window.navigator.language.

https://github.com/appcelerator/alloy/blob/master/Alloy/builtins/moment.js#L894-L914

So it looks like the existing `moment.lang(Ti.Locale.currentLocale);` is OK. Or does your test show otherwise?

mcvendrell:
In my tests what I've found is that `moment.lang(Ti.Locale.currentLocale);` retrieves `es-ES` value and don't works in my project (I see all moments in english), while setting `moment.lang(Ti.Locale.currentLanguage);` retrieves `es` and works perfect (I see spanish).

FokkeZB:
Perhaps we should upgrade momentjs from 2.7 to 2.10.6

falkolab:
Current alloy/moment does not do fallback. Yes, it will be better to upgrade momentjs at any case.

FokkeZB:
Wouldn't this work?

```
moment.lang([Ti.Locale.currentLocale, Ti.Locale.currentLanguage])
```

falkolab:
@FokkeZB  no. This is right code:
```javascript
var moment = require('alloy/moment');
// you MUST require possible language!
require('alloy/moment/lang/ru'); // but Ti.Locale.currentLocale is 'ru-RU'
// ['ru', 'ru-RU']  Moment will be use 'ru-RU' for first.
moment.lang([Ti.Locale.currentLanguage, Ti.Locale.currentLocale]); 
Ti.API.info(moment().format('L')); //08.12.2015 it is right value
```
if you do not use `require` it will be:
```javascript
// run app clean before test
Ti.API.info(moment().format('L')); //12/08/2015 it is wrong value!
```
Look at my exapmle above.

FokkeZB:
Yes of course, you do need to require the languages. But I mean the change in this PR could be:

```
require('alloy/moment/lang/ru');
require('alloy/moment/lang/ru-RU');
moment.lang([Ti.Locale.currentLocale, Ti.Locale.currentLanguage]);
```

mcvendrell:
Not sure about `currentLocale`. It always returns country-COUNTRY and, e.g., ru-RU is not an existent value in `moment/lang` assets.

Not sure if it will work for the few 4-letter assets in `moment/lang` (currently only 9), but if it works for those 9 exceptions, then would be the solution.

falkolab:
@FokkeZB It must be ```[Ti.Locale.currentLanguage, Ti.Locale.currentLocale]``` but not ```[Ti.Locale.currentLocale, Ti.Locale.currentLanguage]```! And yes, file `alloy/moment/lang/ru-RU.js` not exists.

FokkeZB:
@falkolab but in the [source](https://github.com/appcelerator/alloy/blob/master/Alloy/builtins/moment.js#L895) it says:

```
        //pick the language from the array
        //try ['en-au', 'en-gb'] as 'en-au', 'en-gb', 'en', as in move through the list trying each
        //substring from most specific to least, but move to the next array item if it's a more specific variant than the current root
```

So that seems to say you can just do `moment.lang(Ti.Locale.currentLocale)` which should then try `en-gb` first and then `en`?

mcvendrell:
Then something is wrong Fokke, because in my project, setting ONLY `moment.lang(Ti.Locale.currentLocale)` for a device in spanish, all is in english. If it works as you said, must appear in spanish and this is not the case.

falkolab:
@FokkeZB source of problem in https://github.com/appcelerator/alloy/blob/master/Alloy/builtins/moment.js#L873
I can't use ru-RU at first because it throws exception then returns with default language.
But if you set ['ru','ru-RU'], ru will be set as default then it try ru-RU with fail back to default ('ru')

FokkeZB:
Yes, that's very odd indeed. Somethings definitely not right here with the logic in this version.

In the current version it seems to do this correctly and not return the default when a language could not be loaded:

https://github.com/moment/moment/blob/develop/min/moment-with-locales.js#L229-L270

feons:
@mcvendrell, there's a PR pending, https://github.com/appcelerator/alloy/pull/747,  which updates moment.js to version 2.10.6. It should resolve the issue you are facing.

Also, it appears that you haven't signed the [CLA] (https://cla.appcelerator.com/). Please do so before we can review your changes. Thanks again for contributing! :)

mcvendrell:
Hi @feons .

Something weird is happening with CLA. I sign it with Github account but nothing happens. Also I cannot login in the "Manage" page with my Appc Account.

Don't know how to proceed.

FokkeZB:
@mcvendrell  could you try to clean your cookies or use private browsing?

mcvendrell:
Ok, that was the problem. Private browsing fixed it.

feons:
@mcvendrell, we have updated moment.js to version 2.10.6, which should address the issue you were facing. 

I think we can close this PR?

feons:
Close the PR as it has been addressed by #747. 
We can revisit this when necessary.

*Code Strong* 🚀

PR #749, opened 2016-01-06
[ALOY-1346] Controller template should no longer declare args variable

FokkeZB:
https://jira.appcelerator.org/browse/ALOY-1346

### DIFF
--- a/Alloy/template/controller.js
+++ b/Alloy/template/controller.js
@@ -1,1 +1,1 @@
-
+var args = $.args;
--- a/Alloy/template/controller.js
+++ b/Alloy/template/controller.js
@@ -1,1 +1,1 @@
-var args = arguments[0] || {};
+
--- a/Alloy/template/controller.js
+++ b/Alloy/template/controller.js
@@ -1,1 +1,2 @@
+// Arguments passed into this controller can be accessed off of the `$.args` object directly or:
 var args = $.args;


skypanther:
This feature was added (by me) in response to the request of two of your solutions engineers. I would make sure you're not removing something that your customers have implemented and expect.

FokkeZB:
Thanks for the heads up @skypanther! I assume that was before we exposed it via `$.args`? I'll ask SE on their opinion about removing this.

rblalock:
I'm for replacing it with `var args = $.args;`.  Leaving that stubbed out is helpful for beginners.

FokkeZB:
Agree with @rblalock @feons, mostly because it won't break (copied) code and expectations. Updated the PR.

grantges:
I'm for actually removing the reference - and replacing it with a comment at the top of the controller file.

// Arguments passed into this controller can be accessed off of the '$.args' object. 
//
// Example
//     myProperty = $.args.myProperty

rblalock:
@grantges ewww

FokkeZB:
```
// Arguments passed into this controller can be accessed off of the '$.args' object. 
//
// Example
// myProperty = $.args.myProperty
//
// You can add a class to a proxy via `$.addClass`.
//
// Example
// $.addClass($.someView, "someclass");
//
// ...
```

😜

rblalock:
I wonder how many people  use the stubbed out controllers?  The last time I did was when...it was in beta :-D

rblalock:
It'd be good to have a stubbed out controller in our docs that has all these comments and such.  e.g. https://facebook.github.io/react/docs/reusable-components.html

FokkeZB:
Yes, I'd prefer that instead of commenting just this one feature in the stub.

FokkeZB:
@feons let's go with the change as it is now.

skypanther:
A new controller (`alloy generate controller <name>`) already gives you the following at the top of the file

```
var args = arguments[0] || {};
```

All the original change did was expose those args such they could be used in TSS style if statements. It added `args` as a property of `$`. There's no real value to a dev to have them available as `$.args` vs that var statement other than maybe some convenience for inexperienced devs that don't know what the var statement is doing.

FokkeZB:
@skypanther well, I think having `$.args` as a documented way to get args instead of having to do `var args = arguments[0] || {}` in every controller has value, don't you agree?

skypanther:
yeah, that would be fine by me. Honestly, I had forgotten that was in there and add the `var args...` statement myself every time I create a controller by hand.

FokkeZB:
Same for me :wink:

FokkeZB:
Updated with comment above `var args = $.args`

feons:
*LGTM*

Alloy/template/controller.js
```diff
@@ -1 +1,2 @@
-var args = arguments[0] || {};
\ No newline at end of file
+// Arguments passed into this controller can be accessed off of the `$.args` object directly or:
```
FokkeZB: Mmm... "off of"?

ewieberappc: I also dislike the "off of" wording. Perhaps "by" would read better?

FokkeZB: Or "accessed via" ?

ewieberappc: works for me

FokkeZB:
Changed :D

FokkeZB:
? that didn't go well. Looks like I have two patch-12 branches now. Will resolve tomorrow

FokkeZB:
@ewieberappc somehow I done something wrong with the branch. Here's an additional PR to be merged after this with the wording:

https://github.com/appcelerator/alloy/pull/753

@feons go ahead ;)

PR #750, opened 2016-01-07
[ALOY-1321] iOS: remove unsupported app icons and launch images for new projects

feons:
https://jira.appcelerator.org/browse/ALOY-1321


FokkeZB:
I see you're also removing launch image? The ticket is just about icons and we don't generate launch images so I think they are still needed?

feons:
I removed the unsupported ones. Are they needed?

FokkeZB:
Well, the ticket is only about the icons, since with `DefaultIcon.png` we no longer need these here. But the launch images are still required, at least until we equip our templates with the new storyboard launch files we'll introduce with 5.2 if I'm correct.

feons:
It's a good thinking on the single responsibility principle!

FokkeZB:
Looks good now, but you can remove the 2 iTunesArtwork files as well from:
https://github.com/feons/alloy/tree/ALOY-1321/platforms/ios/project/app/assets/iphone

These are generated (for adhoc only) as well:
https://github.com/appcelerator/titanium_mobile/blob/816d68323c588bd1143c2f9bb8f52dbd65e08255/iphone/cli/commands/_build.js#L4509-L4510

feons:
**Approved.**

PR #754, opened 2016-02-08
[AC-610] add support for xhigh on iOS in measurement

manumaticx:


feons:
Looks good. **Approved!**

PR #756, opened 2016-02-12
[ALOY-1355] Fixed issue where attributes containing 'true' or 'false'…

cb1kenobi:
… where being treated as strings instead of booleans. There's no way for Alloy to know if the attribute really is a boolean or not, so we have to blindly test for 'true' and 'false'. This isn't ideal. If someone has a label of 'true', then it will be cast as a bool, but *should* be cast back to a string when used in the context of a string value.

### DIFF
--- a/Alloy/commands/compile/compilerUtils.js
+++ b/Alloy/commands/compile/compilerUtils.js
@@ -225,6 +225,11 @@ exports.getParserArgs = function(node, state, opts) {
 					createArgs[CONST.CLASS_PROPERTY] = theValue.split(/\s+/) || [];
 				}
 			} else {
+				if (theValue === 'true') {
+					theValue = true;
+				} else if (theValue === 'false') {
+					theValue = false;
+				}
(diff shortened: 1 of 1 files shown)


FokkeZB:
Actually, same is true for numbers. I you do `<TabbedBar index=0>` it will be a `"0"` string currently, while it should clearly be a `0` Number.

cb1kenobi:
@FokkeZB Fixed. :)

skypanther:
Last I checked, if you used something like `<View visible=false/>` Alloy would output a warn in the logs. If that is going to be acceptable XML, the logging should be updated to accommodate it. Maybe that's from xmldom?

FokkeZB:
@skypanther it is coming from xmldom.

See https://jira.appcelerator.org/browse/ALOY-1355?focusedCommentId=376959&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-376959

feons:
Looks good. **Approved.**

PR #762, opened 2016-02-25
Added check for variable existence before calling off in destroy function

brentonhouse:
If you have an Alloy component that uses databinding and has conditional code defined in the xml, the destroy function is not taking into consideration the fact that the component may be undefined before calling the `off` function.


FokkeZB:
The beauty is in the small things.

@feons code looks good, CLA server is still down :(

feons:
**Approved.**

PR #767, opened 2016-03-16
[ALOY-1368] Complex data binding for individual model doesn't work with transform() method

FokkeZB:
https://jira.appcelerator.org/browse/ALOY-1368


FokkeZB:
I will include this in the fix for https://jira.appcelerator.org/browse/ALOY-1474 since it will conflict.

PR #768, opened 2016-03-16
[ALOY-1254] Alloy: Support model transform() method for collection-view binding

FokkeZB:
https://jira.appcelerator.org/browse/ALOY-1254


feons:
Looks good. **Approved.**

PR #769, opened 2016-03-16
[ALOY-1369] Complex data binding for collection doesn't work with transform

FokkeZB:
https://jira.appcelerator.org/browse/ALOY-1369


FokkeZB:
@feons, as you can see a few tests are failing. This is because of the following:

* [generateStyleParams](https://github.com/appcelerator/alloy/pull/769/files#diff-a59a2178e486f81366775c39e5393695R570) now returns a string that includes tpl vars, e.g.:

````
      title: _.template('<%=title%>', __alloyId4.__transform),
      enabled: _.template('<%=enabled%>', __alloyId4.__transform),
```

Which then is sometimes used in another `_.template()` method in e.g. with [ItemContainer](https://github.com/appcelerator/alloy/blob/master/Alloy/commands/compile/parsers/Alloy.Abstract._ItemContainer.js#L90-L92).

And since the variables are not found in the settings passed to the template call, the compile crashes like:

```
((__t=(title))==null?'':__t)+
       ^
ReferenceError: title is not defined
```

Any idea how we can best handle this?

FokkeZB:
@feons I solved it by using a different template notation in the styler. All tests pass (locally, running on Travis now)

feons:
Looks good. **Approved.**

FokkeZB:
Conflict solved, tests running

PR #772, opened 2016-03-23
[TIMOB-20085] "ti clean" does not scan SDK hooks

infosia:
Related to [TIMOB-20085](https://jira.appcelerator.org/browse/TIMOB-20085) and https://github.com/appcelerator/titanium_mobile/pull/7628

[hooks/deepclean.js](https://github.com/appcelerator/alloy/blob/master/hooks/deepclean.js) prevents from running SDK-specific `ti clean` hook because `deepclean.js` doesn't properly call finish callback.

### DIFF
--- a/hooks/deepclean.js
+++ b/hooks/deepclean.js
@@ -12,9 +12,10 @@ exports.init = function (logger, config, cli, appc) {
 		fs = require('fs'),
 		afs = appc.fs;
 
-	function run() {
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
:+1: LGTM. Need to speak with @eric34 and @cheekiatng about when how to merge so this only gets in for 5.3.0, not 5.2.1.

PR #773, opened 2016-03-23
[ALOY-1365] Add i18n and platform to default gitignore

FokkeZB:
https://jira.appcelerator.org/browse/ALOY-1365

### DIFF
--- a/Alloy/template/gitignore.txt
+++ b/Alloy/template/gitignore.txt
@@ -1,4 +1,6 @@
 .DS_Store
+i18n
+platform
 Resources
 build.log
 build


cb1kenobi:
How many templates does Alloy need?

I fixed it and I added / to each directory as to not ignore those directories if they're in another directory: https://github.com/cb1kenobi/alloy/commit/add2c60c7cbdb800b4d66e6ab515cdcabd23cfc7.

FokkeZB:
It needs less.

I'm all for removing the Alloy specific icons and splashes as well as the builtin templates and simply have a few simple Titanium templates, all but maybe one using Alloy.

cb1kenobi:
It's on the radar for Alloy v2.

PR #776, opened 2016-03-28
[ALOY-1474] [ALOY-1368] Fixes regression in model data binding without model transform() and bug with

FokkeZB:
https://jira.appcelerator.org/browse/ALOY-1474


FokkeZB:
If you run the test app it should look like the following to demonstrate model and collection binding, simple and complex, without and with model transformer method, with and without custom controller transform method.

simulator screen shot 28 mar 2016 10 21 32

FokkeZB:
@feons ignore the failing travis-ci/push hook

feons:
Looook great! **Approved.**

PR #791, opened 2016-06-09
[ALOY-1503] After Alloy compile, property which has array disappear.

feons:
https://jira.appcelerator.org/browse/ALOY-1503


cheekiatng:
APPROVED.

PR #792, opened 2016-06-29
[ALOY-1508] iOS: Localised splash screen is not visible

feons:
https://jira.appcelerator.org/browse/ALOY-1508


cheekiatng:
CR and FT passed!

PR #793, opened 2016-06-30
[TIMOB-20092] iOS: Alloy Introducing Snapshotter

AngelkPetkov:
[JIRA](https://jira.appcelerator.org/browse/TIMOB-20092?filter=-1)


hansemannn:
Since it's no UI-Element, we don't need this change. PR declined.

PR #799, opened 2016-09-22
[ALOY-1519]  Support iOS 10 <RefreshControl> tag in Ti.UI.ScrollView

feons:
https://jira.appcelerator.org/browse/ALOY-1519


hansemannn:
LGTM, approved! Thank you @feons 🙂

hansemannn:
Approved! @feons: The fix-version is set to 6.0.0, can we ensure this fix goes in? Which Alloy version would that be? Thx!

PR #800, opened 2016-10-26
[ALOY-1526] Special text property value is not parsed correctly

feons:
https://jira.appcelerator.org/browse/ALOY-1526


cheekiatng:
CR and FT passed!

PR #801, opened 2016-10-26
[ALOY-1523] Update Alloy old require-paths

feons:
https://jira.appcelerator.org/browse/ALOY-1523


hansemannn:
Not required anymore, since we are allowing require-backwards-compatibility in 6.0.0.GA

hansemannn:
@topener Can you compare the results?

feons:
@hansemannn, @topener, are we ready to merge this?

Topener:
It seems to do exactly what I did in my code to improve. Especially the 3 `create` functions in the global `Alloy` namespace speed it up a lot

hansemannn:
@feons LGTM, but I don't feel I'm the appropriate person to review. Can @sgtcoolguy take a peek, since he did the require changes in the core? And @Topener, did you test this specific PR in your real app? That would help tremendously!

sgtcoolguy:
So this LGTM, but I should note that this may *break* the code on Titanium < 6
May want to test and see if this breaks on Ti 5.x and if we care? If it doesn't break it, or we don't care, then fine to merge. Otherwise you'll have to toggle the leading slash prefix based upon the SDK version.

feons:
@sgtcoolguy, I was looking at [JavaScript Module Path Resolution](https://wiki.appcelerator.org/display/guides2/CommonJS+Modules+in+Titanium), the changes shouldn't break on Ti 5.x since 

> if the path is prefixed with a /, the module path is also resolved relative to the Resources directory

hansemannn:
Should we also update all the wikis / Alloy docs?

sgtcoolguy:
@feons I see that's what our docs say, but did you actually try it out? It's been known to happen that docs don't match reality 😏

PR #802, opened 2016-11-01
[ALOY-1527] Cannot use Ti.UI.iOS.TabbedBar in ListView template

feons:
https://jira.appcelerator.org/browse/ALOY-1527


hansemannn:
PR approved!

PR #803, opened 2016-11-02
support multiple restrict files

yomybaby:
#625 is a PR to allow compile only one specific file.
This PR supports compiling several specific files.

example:

```shell
# file=view1.js,file=view2.js,file=view3.js ...

alloy compile --config platform=ios,file=app/controllers/index.js,file=app/controller/myCtrl.js
```

### DIFF
--- a/Alloy/commands/compile/index.js
+++ b/Alloy/commands/compile/index.js
@@ -71,7 +71,11 @@ module.exports = function(args, program) {
 		logger.debug('raw config = "' + program.config + '"');
 		_.each(program.config.split(','), function(v) {
 			var parts = v.split('=');
-			alloyConfig[parts[0]] = parts[1];
+			if(alloyConfig[parts[0]]){
+				alloyConfig[parts[0]] = [].concat(alloyConfig[parts[0]],parts[1]);
+			}else{
+				alloyConfig[parts[0]] = parts[1];
+			}
 			logger.debug(parts[0] + ' = ' + parts[1]);
 		});
 	}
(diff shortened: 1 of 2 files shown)


gimdongwoo:
Love it!
Very good!!

hansemannn:
Same as #805, please file a ticket.

Sophrinix:
You beat me to implementing this feature. I'm Very happy to see this! Please tell me this will work outside the app directory.

yomybaby:
I can not crate alloy project issue. So I just create issue on Titanium JIRA project.
https://jira.appcelerator.org/browse/TIMOB-24172

yomybaby:
@Sophrinix What do you mean `work outside the app directory`? Could you please more detail, please? 
"compile file which is not in the app directory" or "run command form outside of app directory"?

feons:
**Approved.**

Sophrinix:
@yomybaby Here's are two use cases for what I am hoping this can accomplish.

Consider the following directory structure for the following example. We are assuming I am using iPhone in this example, although replacing iPhone with android makes no overall difference.
I have taken liberties to collapse certain directories for space:

```
$ tree -d app

├── assets
├── controllers  
├── i18n 
├── lib  
├── migrations
├── models
├── platform
├── styles
├── themes
│   ── TeamA
│       ├── alloy
│       │   └── controllers
│       │    └── example.js
│       │   └── views
│       │    └── example.xml
│       ├── assets
│       └── styles
├── views
└── widgets
    
```

Currently, if you run alloy compile, the views inside ```/app/views``` will be transformed and appended to the controllers. However, before that happens the files in ```themes/TeamA/alloy/*``` will be recursively copied into ```Resources/iphone/alloy/*``` They will however not be compiled/transformed. If the alloy compile command could be used to compile these files in ```themes/TeamA/alloy/``` it would vastly improve the flexibility and power of themes, as well as alloy. This same trick works if you place these files in ```/app/lib/alloy```.

Why would you want to place these files in ``` lib```? git submodules offer one possible reason. Let's say you have functionality that is reusable and you don't want to make it into a widget for reasons outside the scope of this example. Since these files are being copied into the ```Resources``` directory anyway, it would be super helpful if they could go ahead and be compiled. 

I was already beginning work on this, but I am hoping you already got there first. 
The worst case scenario idea would be that we would use this command inside ```alloy.jmk```

PR #804, opened 2016-11-04
add 'skipOptimize' option to alloy compile config

yomybaby:
## Problem
When alloy compiling, all js files be optimized.
If you project has many js lib, optimizing take quite long time.

## Solution
This PR make you can skip js file optimizing manually.
If you want to skip optimization when you alloy compile, add `skipOptimize=true` option to alloy config:
```
alloy compile --config platform=ios,skipOptimize=true
```

You can save time using this options.

On my MBP,
Alloy compiling of default alloy project took 0.9s without skipOptimize.
But it took 0.3s with skipOptimize. 👍

### DIFF
--- a/Alloy/commands/compile/index.js
+++ b/Alloy/commands/compile/index.js
@@ -446,6 +446,8 @@ module.exports = function(args, program) {
 
 	if (restrictionSkipOptimize) {
 		logger.info('Skipping optimize due to file restriction.');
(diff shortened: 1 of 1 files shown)


yomybaby:
This PR caused a side effect.
I'll close this PR.
After fix problem, I'll reopen this.

PR #808, opened 2016-11-14
[ALOY-1517] UI elements are not created correctly when using conditio…

feons:
…nal styles
https://jira.appcelerator.org/browse/ALOY-1517


feons:
**Approved.**

PR #814, opened 2017-02-01
[ALOY-1540] npm node_modules are erroring when compiling

feons:
https://jira.appcelerator.org/browse/ALOY-1540


Bertuzb:
does this also work on nested `node_modules`? i.e. `app/lib/mymodule/node_modules/**`. I dont see an immediate usecase for this but it gives the developer a bit more freedom.

feons:
@Bertuzb, good point!

feons:
**Approved.**

PR #815, opened 2017-02-07
Moved creating index controller from generation to template

Bertuzb:
https://jira.appcelerator.org/browse/AC-4777

Solution shouldnt be much more than this


feons:
@Bertuzb, I think this breaks existing apps that startup from index.

Bertuzb:
yes when they update it breaks, but this is a much better solution instead of appending unknown code during the compilation process.

Fixing it for any app will be really simple though, simply append `Alloy.createController('index')` as the last line on alloy.js and you're good to go

PR #818, opened 2017-02-24
Add optimizingExeptions option to config.json

yomybaby:
This PR make you can skip the file when doing an optimization manually.
There are no reason to optimize js files which doesn't depends on alloy.

If you want to skip, add `optimizingExeptions` to `config.json`:
- If `app/lib/q.js`, the path should be `q.js`(see below. After compiling, `app/lib/q.js` to `q.js`)

```javascript
{
  //...
  "optimizingExceptions": [
        "moment2.11.1.js",
        "q.js",
        "underscore1.8.3.js",
        "sf.js",
        "nl.fokkezb.html2as.js"
    ],
  //...
}
```

### DIFF
--- a/Alloy/commands/compile/index.js
+++ b/Alloy/commands/compile/index.js
@@ -1107,7 +1107,8 @@ function optimizeCompiledCode(alloyConfig, paths) {
 			'alloy/constants.js',
 			'alloy/underscore.js',
 			'alloy/widget.js'
-		];
+		].concat(compileConfig.optimizingExceptions || []);
+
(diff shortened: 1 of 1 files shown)


feons:
@yomybaby, like the idea! Could you please create a ticket and also fix the conflict. Thanks!

yomybaby:
All done.
https://jira.appcelerator.org/browse/AC-4825

Thanks!

feons:
**Approved.**

yomybaby:
🙏🏻

PR #821, opened 2017-03-24
[AC-4875] Update gitignore

m1ga:
JIRA-Ticket: https://jira.appcelerator.org/browse/AC-4875

```
ti create -t app -n test --id com.test.test -p all -u test.de -d .
cd test
alloy new
```
still had an old gitignore without the /platform and /i18n (like in https://github.com/appcelerator/alloy/blob/master/templates/default/.gitignore)

### DIFF
--- a/Alloy/template/gitignore.txt
+++ b/Alloy/template/gitignore.txt
@@ -1,5 +1,7 @@
 .DS_Store
-Resources
+/Resources
+/platform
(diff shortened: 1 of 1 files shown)


hansemannn:
LGTM! @feons any objections?

feons:
Looks good! **APPROVED**

PR #823, opened 2017-04-18
[ALOY-1263] Enable strict mode in controller

feons:
https://jira.appcelerator.org/browse/ALOY-1263

Enabling this in new projects is easy, enforcing this on an existing project might require a bit more work from developers depending on how sloppy the controllers are.

To opt into strict mode, add `"strict": true` in `config.json`.

### DIFF
--- a/Alloy/commands/compile/index.js
+++ b/Alloy/commands/compile/index.js
@@ -571,7 +571,8 @@ function parseAlloyComponent(view, dir, manifest, noView, fileRestriction) {
 				fs.readFileSync(path.join(alloyRoot, 'template', 'wpath.js'), 'utf8'),
 				{ WIDGETID: manifest.id }
 			),
-			__MAPMARKER_CONTROLLER_CODE__: ''
+			__MAPMARKER_CONTROLLER_CODE__: '',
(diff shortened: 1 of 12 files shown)


hansemannn:
I think @cb1kenobi should rather review this.

cb1kenobi:
So, how much performance improvement does this PR help Android apps?

feons:
@cb1kenobi, according to the ticket description, Fokke managed to get a 20x performance increase.  I think this is best answer by the large number of developers pushing for this feature.

cb1kenobi:
@feons That means nothing. The ticket is 2 years old. V8 has been updated since then. Here's the app Fokke used to test: https://github.com/appcelerator-developer-relations/Mobile-App-Performance. How much of a performance improvement does this PR help?

feons:
@cb1kenobi, frankly I personally see no performance gain in running this using some of the alloy samples. The main motivation for this PR to allow developers the to opt in strict mode more easily;  catches some common coding bloopers; be more stringent and prevents relatively “unsafe” actions.

hansemannn:
Let's not take this PR then, as it was based on outdated benchmarking tests. Closing this PR and resolving the linked ticket for now, sorry for the possible confusion!

PR #827, opened 2017-05-26
[ALOY-1567] Removed minification code

cb1kenobi:
JIRA: https://jira.appcelerator.org/browse/ALOY-1567


sgtcoolguy:
This won't work unless we're agreeing to migrate behavior to the main SDK magnification - and it will break there if we do anyways.

Those 3 babel plugins were the equivalent of the old uglifyjs compressor stuff that alloy relies on to 'optimize' dead code after inserting booleans for the OS_* constants and string values for Ti.Platform.name and Ti.Platform.osname. The unit tests assume they will run and prune dead code here.

I've argued that this behavior should likely move over to the main SDK build process anyways, since it is generally useful. But this does break the tests as-is and will change alloy's behavior.

Also, a note that even if we remove babili and/or those plugins here, they'll still get run in the SDK when we *are* doing minification - which I'm assuming will generate "broken" code for TiCore?

cb1kenobi:
The minification should ONLY be done by Titanium since Titanium controls whether we should minify in the first place.

As far as dead code removal is concerned, I'm fine with this in Alloy, however `babel-plugin-minify-dead-code-elimination` has a bug where hoisted functions defined in block scopes are removed instead of being moved to the parent scope. We'll have to mark the tests as 'skip' until `babel-plugin-minify-dead-code-elimination` is fixed.

Yes, the babili code in `node-titanium-sdk` is will also cause issues and we'll need to remove `babel-plugin-minify-dead-code-elimination` from there too.

feons:
**Approved.**

PR #828, opened 2017-05-30
[ALOY-1566] OS_* declaration should not choke compile

feons:
https://jira.appcelerator.org/browse/ALOY-1566


feons:
**Approved.**

PR #829, opened 2017-05-31
fix optimizer tests

feons:
Code minification has been removed from optimising step, some of the tests should be updated.


feons:
Tests passed locally. *Approved.*

PR #830, opened 2017-06-14
[ALOY-1569] gen WPATH function to match updated require-handling in T…

feons:
…itanium 6

https://jira.appcelerator.org/browse/ALOY-1569


feons:
**Approved.**

PR #831, opened 2017-06-16
Add babel-preset-es2015

gimdongwoo:
I am really happy to be babel applied to the alloy.

Thank you very much to @sgtcoolguy.

I think ES2015 (ex: arrow functions, const, let) are already used as standard, so we need es2015 preset.

Thank you.

### DIFF
--- a/Alloy/commands/compile/sourceMapper.js
+++ b/Alloy/commands/compile/sourceMapper.js
@@ -24,7 +24,7 @@ exports.OPTIONS_OUTPUT = {
 	comments: false,
 	babelrc: false,
 	passPerPreset: false,
-	presets: ['es2015']
+	presets: [require('babel-preset-es2015')]
 };
 
 function mapLine(mapper, theMap, genMap, line) {
--- a/Alloy/commands/compile/sourceMapper.js
+++ b/Alloy/commands/compile/sourceMapper.js
@@ -23,7 +23,8 @@ exports.OPTIONS_OUTPUT = {
 	compact: false,
 	comments: false,
 	babelrc: false,
-	passPerPreset: false
+	passPerPreset: false,
+	presets: ['es2015']
 };
 
 function mapLine(mapper, theMap, genMap, line) {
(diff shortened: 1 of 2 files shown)


brentonhouse:
Why would you want to transpile to es5?  Why not leave it as es6?

gimdongwoo:
@brentonhouse Because iOS does not support es2015's many things (ex: let, const and etc) even if you use jscore.

brentonhouse:
iOS seems to have really good compatibility w/ es6:  https://kangax.github.io/compat-table/es6/

gimdongwoo:
I'm sorry. I did not mention iOS8 & 9 by mistake. Titanium SDK 6.1.0.GA also supports iOS8 & 9.

The smart way is to determine the transcoding according to the ios version specified in tiapp.xml.
I will update this PR by adding this part.

sgtcoolguy:
@gimdongwoo  I see you closed this PR, but I like the idea in general. However, I don't think Alloy is the right place to do this. In an alloy project we basically do two build "passes". First alloy generates code and does some optimizations, etc sticking the result into the `Resources` folder, which the "classic" titanium CLI/project uses and builds. I think that the transpilation needs to occur in that second step. We should look at the combination of whether the tiapp is using TiCore or JSCore and what minimum iOS version it targets to determine what preset/transpiling needs to occur (if any).

Right now, the babel parsing/minification process is done here: https://github.com/appcelerator/node-titanium-sdk/blob/master/lib/jsanalyze.js#L111

It's called from here for iOS: https://github.com/appcelerator/titanium_mobile/blob/master/iphone/cli/commands/_build.js#L5681

I'd say we should pass along some options that determine the target presets/plugins to run (in addition to minification if we choose that) based on the analysis of the target iOS/JS engine. Then when we do our final pass through babel we apply it all there.

brentonhouse:
@sgtcoolguy -- It would be very cool if was just a configuration (global and per project) that would allow you to define the babel presets you wanted to run.  That way you could run any combination of presets for your project.  The extensibility provided by that would be awesome!

PR #832, opened 2017-06-20
[ALOY-1570] Alloy: exports.baseController does not work since 1.10.0 (Regression)

sgtcoolguy:
*JIRA* https://jira.appcelerator.org/browse/ALOY-1570

*Description*
This is a typo when I converted the code from Uglify to Babel. It used to look for `exports.baseController` and I fudged the translation to be `export.baseController`.

### DIFF
--- a/Alloy/commands/compile/ast/controller.js
+++ b/Alloy/commands/compile/ast/controller.js
@@ -3,7 +3,7 @@ var U = require('../../../utils'),
 	types = require('babel-types'),
 	traverse = require('babel-traverse').default;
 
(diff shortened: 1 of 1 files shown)


feons:
@sgtcoolguy, now we are getting the correct controller name, but it's not a string.

`	require('/alloy/controllers/' + Base).apply(this, Array.prototype.slice.call(arguments));
`
should be 
`	require('/alloy/controllers/' + "Base").apply(this, Array.prototype.slice.call(arguments));
`

feons:
*LGTM*

PR #834, opened 2017-06-20
[ALOY-1569] gen WPATH function to match updated require-handling in T…

feons:
…itanium 6 1_9_X

https://jira.appcelerator.org/browse/ALOY-1569


feons:
**Approved.**

PR #836, opened 2017-06-23
Fixing exceptions to include others app.js

kopiro:
https://jira.appcelerator.org/browse/AC-5054

If your Alloy project has an `app.js` in a subdirectory (example: `lib/app.js`), it is not compiled.
So, if you have a pre-processor constant like `OS_IOS` or `OS_ANDROID` in this file,
the result is a runtime error because no var defined `OS_IOS` is found.

*This is a critical issue, because Alloy produces a broken app!*

### DIFF
--- a/Alloy/commands/compile/index.js
+++ b/Alloy/commands/compile/index.js
@@ -1094,32 +1094,42 @@ function optimizeCompiledCode(alloyConfig, paths) {
 			return fileRestrictionUpdatedFiles;
 		}
 
-		var exceptions = [
+		var singleFileExceptions = [
 			'app.js',
 			'alloy/CFG.js',
-			'alloy/controllers/',
-			'alloy/styles/',
 			'alloy/backbone.js',
 			'alloy/constants.js',
 			'alloy/underscore.js',
-			'alloy/widget.js',
+			'alloy/widget.js'
+		];
+
+		_.each(singleFileExceptions.slice(0), function(ex) {
+			singleFileExceptions.push(path.join(titaniumFolder, ex));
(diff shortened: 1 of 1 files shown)


Alloy/commands/compile/index.js
```diff
@@ -1094,32 +1094,42 @@ function optimizeCompiledCode(alloyConfig, paths) {
...
 		});
 
 		var rx = new RegExp('^(?!' + otherPlatforms.join('|') + ').+\\.js$');
+
 		return _.filter(wrench.readdirSyncRecursive(compileConfig.dir.resources), function(f) {
-			return rx.test(f) && !_.find(exceptions, function(e) {
-				return f.indexOf(e) !== -1;
-			}) && !fs.statSync(path.join(compileConfig.dir.resources, f)).isDirectory();
+			var test = rx.test(f) && 
+			(!_.find(directoryExceptions, function(e) { return f.indexOf(e) !== -1; }) ||
```
feons: Can we do something like the following, so that we can remove the block of `_.find` calls?
```
var exceptions = [
    ...
    '.+node_modules'
].concat(compileConfig.optimizingExceptions || []);
...
var excludePatterns = exceptions.concat(otherPlatforms || []);
var rx = new RegExp('^(?!' + excludePatterns.join('|') + ').+\\.js$');
...
    return rx.test(f) && !fs.statSync(path.join(compileConfig.dir.resources, f)).isDirectory();
```

feons:
Close in favor of https://github.com/appcelerator/alloy/pull/839

PR #841, opened 2017-08-09
Move Toolbar to Ti.UI.Toolbar namespace

ypbnv:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-25003

**Description:**
Remove implicit iOS namespace for Toolbar.
Add Android alloy parser for Toolbar.

I believe this would keep the backward compatibility for iOS since the new code is used only when the Toolbar is used as a custom ActionBar (which we will alloy only for Android). If we can further guard for that we could add that as well.

### DIFF
--- a/Alloy/commands/compile/parsers/Ti.UI.Toolbar.js
+++ b/Alloy/commands/compile/parsers/Ti.UI.Toolbar.js
@@ -21,7 +21,7 @@ exports.parse = function(node, state) {
 	var tempRes = require('./Alloy.Abstract._ItemContainer').parse(node, state);
 
 	//Only if the toolbar is passed as an ActionBar to the activity
-	if (state.parent.node.getAttribute('customToolbar') === node.getAttribute('id')) {
+	if (state.parent && state.parent.node && state.parent.node.getAttribute('customToolbar') === node.getAttribute('id')) {
 
 		var activityTssStyles = _.filter(state.styles, function(elem) {
 			// generates a sorted array of styles filtered to include only elements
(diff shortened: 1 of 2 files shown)


feons:
@ypbnv, please also update the samples and codes that reference `Ti.UI.iOS.Toolbar`, also `Ti.UI.iOS.Toolbar.js` should be removed.

hansemannn:
@feons Can we keep it until 8.0.0? Otherwise it will be a breaking change for Alloy, although we still support the tag. It's deprecated in 6.2.0 and will be removed in 8.0.0.

hansemannn:
@ypbnv Testing today. Does Alloy throw a deprecation-warning as well? Not sure if it's required though, since the SDK throws warnings already.

ypbnv:
@hansemannn I don't know. I didn't even think about deprecation warnings, because it is something new for Android.

hansemannn:
I mean the Ti.UI.iOS namespace, but that's fine then. Will check it out!

ypbnv:
Can I even access anything in that namespace? I have been building only for Android.

feons:
@hansemannn, sure. Actually I thought this is supposed to be a backward compatible changes?

feons:
ops.. closed by accident ...

hansemannn:
@feons Is the fix-version okay as it is?

feons:
@hansemannn,

- could you please create an alloy ticket for this?
- what fix-version are you referring to?
- most importantly, there are quite a number of references to `Ti.UI.iOS.Toolbar` in the code base that are not taken care of, i.e sample code with `Ti.UI.iOS.Toolbar` usage does cause confusion.

hansemannn:
@feons Sorry for merging too early then! Should I create a ticket? The old toolbar still works for me, but please do another test to verify, that'd be awesome!

feons:
https://jira.appcelerator.org/browse/ALOY-1578 to track updates.

PR #843, opened 2017-08-22
Bring in lodash.mergeWith, use to merge the babel plugins

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-24980

Bring in lodash.mergewith (from npm package for this specific method), use to merge the babel plugins used to transform the code.

I think this may need some user testing, specifically from @brentonhouse to see if this even will achieve what he's after. plugin and preset ordering has meaning for babel, so it may be difficult to properly allow users to add their own (because they may need to run before or after our own plugins do modifications to the code). This PR only is trying to use the deep merge, not actually add an official config key/value pair for babel plugins (yet).

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -14,6 +14,7 @@ docs/
 **/jasmine.js
 **/lib/colors.js
 **/underscore.js
+**/lodash.mergewith.js
 **/backbone.js
 Alloy/grammar
 Alloy/builtins/social.js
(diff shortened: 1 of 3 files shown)


sgtcoolguy:
cc @brentonhouse

feons:
@brentonhouse, have you try this out? Does it fit your needs?

hansemannn:
This should become obsolete once #842 is merged I *think*?

ewanharris:
From discussions we believe the use case this aimed to achieve to now be possible, so I'm going to close this PR

PR #845, opened 2017-08-30
[ALOY-1579] Support Titanium.UI.Android.DrawerLayout

feons:
https://jira.appcelerator.org/browse/ALOY-1579


mukherjee2:
Passed FR. I tried an example app using Titanium.UI.Android.DrawerLayout on an Android 7, and was able to get the app to work.

PR #846, opened 2017-08-31
[ALOY-1578] Move Ti.UI.iOS.Toolbar to Ti.UI.Toolbar additional cleanup

feons:
https://jira.appcelerator.org/browse/ALOY-1578
Ref: https://github.com/appcelerator/alloy/pull/841


hansemannn:
Looks good! Can people still use the old implementation then? So let's say they use latest Alloy with SDK 6.1.0, not 6.2.0?

feons:
@hansemannn, since #841, alloy will only generated `Ti.UI.createToolbar` regardless sdk version.  Looks like more updates are needed then.

Alloy/commands/compile/parsers/Alloy.Abstract.Items.js
```diff
@@ -4,7 +4,7 @@ exports.parse = function(node, state) {
 	_.extend(state, {
 		itemArrayDefinition: {
 			parents: [
-				'Ti.UI.iOS.Toolbar',
```
hansemannn: Can't we just leave this line and the `Ti.UI.iOS.Toolbar.js` to make it backwards-compatible?

feons: @hansemannn, probably not. The implicit namespace for `Toolbar` is removed, so  `Ti.UI.iOS.Toolbar.js` is not used during compilation.

hansemannn:
@feons There was a report by the community with RC:
```
[ERROR] TiExceptionHandler: (main) [215,215] ----- Titanium Javascript Runtime Error -----
[ERROR] TiExceptionHandler: (main) [1,216] - In /alloy/controllers/index.js:63,37
[ERROR] TiExceptionHandler: (main) [0,216] - Message: Uncaught TypeError: Cannot read property 'createToolbar' of undefined
[ERROR] TiExceptionHandler: (main) [0,216] - Source:     $.__views.__alloyId0 = Ti.UI.iOS.createToolbar({
```

This is probably fixed with this PR?

feons:
@hansemannn, yes. Please use `Ti.UI.createToolbar` in the controller for Android build.

hansemannn:
They only used the Alloy `<Toolbar>` tag, not a manual controller-code. So either it was an old CLI version or this PR takes care of it.

PR #848, opened 2017-09-01
[ALOY-1579] Support Titanium.UI.Android.DrawerLayout 1_9_X

feons:
https://jira.appcelerator.org/browse/ALOY-1579


feons:
Test passed locally. **Approved.**

PR #849, opened 2017-09-17
[TIMOB-25269] iOS: Add iPhone X screenshots

hansemannn:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-25269

To be required in SDK 6.3.0+


hansemannn:
@feons Should I file an own Alloy Ticket for the CI? Still not sure how the Alloy CI magic works 😙.

feons:
@hansemannn, we can track this through existing TIMOB ticket. For CI magic, as as all other CLI modules, you'll need to bump package.json version.

hansemannn:
Ok. So per semantic versioning, it would be 1.11.0, but could we actually deliver that latest with SDK 6.3.0 / CLI 6.x ?

PR #851, opened 2017-09-21
[ALOY-1582] Fix Ti.Platform.osname on Windows

ewanharris:
JIRA:  https://jira.appcelerator.org/browse/ALOY-1582

`Ti.Platform.osname` should report windowsstore when target is ws-local, and windowsphone when target is wp-emulator/wp-device


feons:
**Approved.**

PR #852, opened 2017-09-28
Update underscore to the latest (1.8.3) version. 

Topener:
I've been running a production app (Roamler) since February, since I've missed a couple new features in underscore, without any problems. It is being used by 1000's of people and so far nothing was encountered. I figured everyone could really benefit from new underscore functionality.

Ticket for it: https://jira.appcelerator.org/browse/AC-5243


brentonhouse:
I think the other PR might take care of this:  https://github.com/appcelerator/alloy/pull/842

Topener:
@brentonhouse it might... but the migration guide of underscore to lodash is quite extensive, so it is a breaking change.
Better to upgrade it first ;) https://github.com/lodash/lodash/wiki/Migrating

brentonhouse:
I think the biggest breaking change is when underscore changed how templates worked in 1.7.   They, unfortunately, do not use semver so they did not increment the major version.   Lodash simply matched their change to maintain compatibility and incremented their major version properly.

Topener:
I mean, lodash would be a breaking change for alloy, underscore update wouldn't really. Just for example the `_.findWhere` method in underscore doesn't exist in lodash.

brentonhouse:
Does Alloy databinding no longer use the _.template function?  Also, for some reason, the Alloy build process shares the same version of underscore and the runtime, so we would need to make sure it doesn't break anything in the compile process.  (i.e. _.template)

Topener:
I don't know if it uses it or not, but as far as I read they reverted the breaking changes of 1.7 into 1.8 regarding templates.

brentonhouse:
Pretty sure they didn't.  Both lodash and underscore now use the "new" method of templating.

feons:
From [underscorejs changelog:](http://underscorejs.org/#changelog)
> 1.7.0
Underscore templates no longer accept an initial data object. _.template always returns a function now.

So additional changes to the Alloy compiler is required to get the update working.

Topener:
@feons keep in mind... it is the underscore included in the app you can use in the app. I have quite a big app using quite a bit of underscore and have it running 1.8.3 since February already. No issues

brentonhouse:
The thing is, I think the `underscore.js` file is currently shared by both the runtime and the compiler.

feons:
@brentonhouse, that's right.
@Topener, with the change, I'm getting following compilation error with `alloy compile` of default apps:

```
TypeError: getTextFromGenerator(...).split is not a function
    at Object.exports.generateCodeAndSourceMap (/Users/feonsua/.nvm/versions/node/v7.10.0/lib/node_modules/alloy/Alloy/commands/compile/sourceMapper.js:69:79)
    at parseAlloyComponent (/Users/feonsua/.nvm/versions/node/v7.10.0/lib/node_modules/alloy/Alloy/commands/compile/index.js:942:15)
    at /Users/feonsua/.nvm/versions/node/v7.10.0/lib/node_modules/alloy/Alloy/commands/compile/index.js:445:6
    at Function._.each._.forEach (/Users/feonsua/.nvm/versions/node/v7.10.0/lib/node_modules/alloy/Alloy/lib/alloy/underscore.js:153:9)
    at /Users/feonsua/.nvm/versions/node/v7.10.0/lib/node_modules/alloy/Alloy/commands/compile/index.js:434:6
    at Function._.each._.forEach (/Users/feonsua/.nvm/versions/node/v7.10.0/lib/node_modules/alloy/Alloy/lib/alloy/underscore.js:153:9)
```

Topener:
@brentonhouse @feons ah yeah that dampens the mood a bit... I didn't know they were shared. When just bundled in the app 1.8.3 works awesome, so it is just the compiling that is broken now...

hansemannn:
We are working on Alloy 2.0.0 now, which will use #842 for a modern underscore replacement. Closing this one.

PR #858, opened 2017-11-09
[ALOY-1590] OS_IOS assigned to a variable causes a crash

feons:
https://jira.appcelerator.org/browse/ALOY-1590


ewieberappc:
FR Passed. Error is no longer present when using OS_IOS, OS_ANDROID, and OS_WINDOWS in variable assignment.

PR #860, opened 2017-12-07
[TISTUD-8942] Android Debugger: Some breakpoints not hit if "Resume" is done anytime during debug

sgtcoolguy:
**JIRA**: https://jira.appcelerator.org/browse/TISTUD-8942

**Description**
This boiled down to "some breakpoints were never honored in alloy apps".

After a lot of digging to isolate where the root cause was, I was able to validate the the debugger/SDK seemed fine, as a built alloy app could be debugged not he generated sources just fine.
Classic apps were OK, so it had to be alloy-specific.

I then dug into alloy's source mapping, since it's really a two-pass operation. First, we generate an initial file and our own source map. Then we hand that off to babel and it's supposed to transform the file more and generate the final source mapping. Great, right? Except babel produces intermittently invalid source maps: babel/babel#6008

The workaround here is to avoid trying to feed source maps to babel or consume the ones they generate, but to have it "retainLines" so our original source map is still correct/valid for line mappings (it may get wonky with columns, but Studio is incapable of handling sub-line precision right now anyways).

### DIFF
--- a/Alloy/commands/compile/sourceMapper.js
+++ b/Alloy/commands/compile/sourceMapper.js
@@ -105,9 +105,12 @@ exports.generateCodeAndSourceMap = function(generator, compileConfig) {
 		]
 	});
 	if (compileConfig.sourcemap) {
-		options.sourceMaps = true;
-		options.sourceMapTarget = target.filename;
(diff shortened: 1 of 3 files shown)


feons:
**Approved!**

PR #863, opened 2017-12-18
[ALOY-1595] alloy compile broken on node 9.3.0

feons:
https://jira.appcelerator.org/browse/ALOY-1595

Fixed in https://github.com/abbr/deasync/commit/3638d542b6c0fece8d7ddf2a02d9bcd8b1109bcb

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -392,12 +392,12 @@
       }
     },
     "deasync": {
-      "version": "0.1.10",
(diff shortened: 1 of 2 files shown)


ewanharris:
@feons I'd like to propose #864 as an alternative as I'd rather see us move away from deasync due to the usage of node internals. I ran through the tests locally and some sample apps locally and saw not much difference in performance etc.

feons:
Close in favor of #864.

PR #864, opened 2017-12-19
[ALOY-1595] Use transformFileSync and remove deasync

ewanharris:
https://jira.appcelerator.org/browse/ALOY-1595


ewanharris:
@sgtcoolguy If you have time reckon you could take a look? I requested a review as I know you wrote this code initially and wondered if there was any specific reason for using the deasync module vs the transformFileSync function?

sgtcoolguy:
Probably because I was a dummy and didn't see the sync variant. I honestly don't recall, but this is much cleaner and removes a dependency, so yay!

PR #868, opened 2018-01-04
[ALOY-1534] Add support for Backbone 1.3.3

feons:
https://jira.appcelerator.org/browse/ALOY-1534

Still not set as default version, due to breaking change in the Backbone library from 0.9 to 1.3, existing apps using backbone@0.9 may not run without modification.


feons:
**APPROVED.**

PR #874, opened 2018-01-10
v1.11.0

ewanharris:
Bump to 1.11.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Alloy Release Notes
 
+---
+### Release 1.11.0
+ * No changes, just a version bump
+
+---
 ### Release 1.10.12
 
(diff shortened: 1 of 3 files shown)


hansemannn:
Can we move the features out of 1.10.12, so devs do not think it's an irrelevant update.

ewanharris:
@hansemannn I don't think we should, 1.10.12 has shipped with those tickets. This just brings the version number in line with the changes we made, so that when we release 7.0.2 CLI (or should it be 7.1.0 with these alloy changes?) the alloy version bump indicates some new features

hansemannn:
In a perfect semantic versioning world, it should be 7.1.0 I guess. @ingo thoughts?

ewanharris:
Oh no, @build has become sentient! :)

PR #880, opened 2018-02-21
[ALOY-1606] Android: Debugger not hitting breakpoints on Windows

feons:
https://jira.appcelerator.org/browse/ALOY-1606

After source map generation, the controller source file got optimized, which invalidates the source map.

### DIFF
--- a/Alloy/commands/compile/index.js
+++ b/Alloy/commands/compile/index.js
@@ -1146,7 +1146,7 @@ function optimizeCompiledCode(alloyConfig, paths) {
 		});
 
 		_.each(exceptions.slice(0), function(ex) {
-			exceptions.push(path.join(titaniumFolder, ex));
+			exceptions.push(`${titaniumFolder}/${ex}`);
 		});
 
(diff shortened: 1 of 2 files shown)


hansemannn:
@sgtcoolguy Can you review this?

Alloy/commands/compile/index.js
```diff
@@ -1146,7 +1146,7 @@ function optimizeCompiledCode(alloyConfig, paths) {
 		});
 
 		_.each(exceptions.slice(0), function(ex) {
-			exceptions.push(path.join(titaniumFolder, ex));
+			exceptions.push(`${titaniumFolder}/${ex}`);
```
ewanharris: `path.posix.join(titaniumFolder, ex)`?

hansemannn:
@feons Can you update your ticket based on @ewanharris' review? I think its the last one remaining for CLI release 7.0.3.

Alloy/commands/compile/index.js
ewanharris: Not gonna block on this, looking at it in the whole context we should be safe to do string concat here for path building

PR #883, opened 2018-03-07
[ALOY-1609] Add plugins/ti.alloy to gitignore

Topener:
**JIRA**: https://jira.appcelerator.org/browse/ALOY-1609

Every time the alloy version changes, the alloy.js plugin file inside the plugin folder changes too. This pollutes git and a default procedure is to remove the file from git alltogether.

### DIFF
--- a/Alloy/template/gitignore.txt
+++ b/Alloy/template/gitignore.txt
@@ -10,3 +10,4 @@ tmp
 .project
 .settings
 Thumbs.db
+plugins/ti.alloy/hooks/alloy.js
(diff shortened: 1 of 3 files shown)


Alloy/template/gitignore.txt
```diff
@@ -10,3 +10,4 @@ tmp
 .project
 .settings
 Thumbs.db
+plugins/ti.alloy/hooks/alloy.js
```
ewanharris: I think we'd want to ignore `plugins/ti.alloy` and not the specific file? There's a couple other files under there that can change (hooks/readme, hooks/deepclean.js, plugin.py)

Topener: yep that actually makes more sense! Will update the PR

PR #886, opened 2018-03-15
Update samples

hansemannn:
This is an ongoing effort to update our Alloy samples to work with the latest environment. I removed the Jasmine tests and migrated the Ti.Includes (removed since years) of the "Advanced > Themes" sample so far, many many more to come. This could eventually be team- and community-driven to speed things up.


hansemannn:
Reminder: Lets move all sample out to own samples. There are thoughts of making an own repo for examples.

PR #891, opened 2018-03-28
[ALOY-1304] Support Node text for TextField and TextArea

feons:
https://jira.appcelerator.org/browse/ALOY-1304


hansemannn:
@feons Can you release this new version, together with #894?

PR #894, opened 2018-04-06
added optional failure callback on alloy/dialogs

joshmocekMovista:
I've added an optional failure callback on alloy/dialogs. This way if 'No' is chosen I can run a function.


Alloy/builtins/dialogs.js
```diff
@@ -51,11 +52,15 @@ exports.confirm = function (args) {
...
 		cancel: 0
 	});
-	alertDialog.addEventListener('click', function (evt) {
+	alertDialog.addEventListener('click', function(evt) {
 		if (evt.index) {
-			if (args.callback) { args.callback(args.evt || {}); }
+			if (args.callback) {
+				args.callback(args.evt || {});
+			}
+		} else if (args.failure) {
```
hansemannn: Not sure if that is the correct naming. I would prefer a `cancel` naming. Otherwise, this should be fine and does not break anything related. @feons for a 2nd opinion.

joshmocekMovista: I changed it @hansemannn & @feons

PR #895, opened 2018-05-04
[ALOY-1617] Attempt to fix SQL error when not using columns in table

hansemannn:
**JIRA**: https://jira.appcelerator.org/browse/ALOY-1617

Although the error is on Android, I think we should fix this in Alloy, since we are using empty column parenthesis `( )` when no columns are specified. iOS may filter this out, but Android complains for a good reason. I did not validate the fix so far, so feedback is welcome!

### DIFF
--- a/Alloy/lib/alloy/sync/sql.js
+++ b/Alloy/lib/alloy/sync/sql.js
@@ -86,6 +86,8 @@ function Migrator(config, transactionDb) {
 		// compose the create query
 		var columns = [];
 		var found = false;
+		var sql;
+
 		for (var k in config.columns) {
 			if (k === this.idAttribute) { found = true; }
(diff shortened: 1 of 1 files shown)


hansemannn:
Did we actually verify the fix by testing the reporters code before merging?

feons:
@hansemannn, feel free to open/revert and assign a QE for FR. 
FYI CR passed.

hansemannn:
Please revert this PR. Android rejects this for a good reason, because it is not valid SQL to create a table without columns. See:
- https://www.w3schools.com/sql/sql_create_table.asp
- https://stackoverflow.com/questions/32961327/create-a-blank-table-with-no-column
- http://sqlfiddle.com/#!9/edd78f (Demo)

PR #904, opened 2018-08-29
[ALOY-1627] Breakpoints not working on Alloy apps

feons:
https://jira.appcelerator.org/browse/TIMOB-19848


lokeshchdhry:
FR Passed.

Breakpoints are hit as expected when there are white spaces in between the lines.

Environment:
Locally built CLI Core.

PR #906, opened 2018-09-12
[ALOY-1620] Replace custom copy file logic with fs-extra copySync call

ewanharris:
https://jira.appcelerator.org/browse/ALOY-1620

Rather than just change the Buffer usage, I just replaced the custom logic with an fs-extra call. I did some fairly unscientific testing and there's no noticeable performance difference


feons:
**Approved.**

PR #908, opened 2018-09-13
[TIMOB-26388] iOS 12: support iPhone XS Max and iPhone XR launch images 

hansemannn:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-26388


Topener:
In that case, go ahead. Didn't know compressing twice worked that well really ;)

PR #910, opened 2018-09-19
[ALOY-1631] Add "theme" CLI parameter

drauggres:
**JIRA:** https://jira.appcelerator.org/browse/ALOY-1631

From [documentation](https://docs.appcelerator.com/platform/latest/#!/guide/Alloy_Themes)

> To use a theme, add it to your Alloy project's config.json file with "theme" as the key and the name of the theme folder as the value

For my application I need to apply different themes for customers, so it would be nice to have ability to define theme from command line.


Topener:
Simple yet great! Why haven't I never thought of this.

PR #911, opened 2018-09-27
fix(runtime): avoid usage of implicit globals

ewanharris:
https://jira.appcelerator.org/browse/ALOY-1633

### DIFF
--- a/Alloy/commands/compile/ast/handle-alloy-globals.js
+++ b/Alloy/commands/compile/ast/handle-alloy-globals.js
@@ -1,0 +1,94 @@
+const toCheck = [ 'Alloy', '_', 'Backbone' ];
+var template = require('@babel/template').default;
+
+var buildRequire = template(`
+	var VARIABLE = REQUIRECALL;
+`);
+module.exports = function(babel) {
+	return {
+		name: 'app.js top level variables global transform',
+		visitor: {
+			CallExpression: function (path, state) {
+				const node = path.node;
+				if (node.callee.name !== 'require') {
+					return;
+				}
+				if (!node.arguments || !node.arguments[0]) {
+					return;
+				}
+				switch (node.arguments[0].value) {
+					case '/alloy':
+						this.imported.push('Alloy');
+						break;
+					case '/alloy/underscore':
+						this.imported.push('_');
+						break;
+					case '/alloy/backbone':
+						this.imported.push('Backbone');
+						break;
+				}
+			},
+			ImportDeclaration (path) {
+				const node = path.node;
+				if (!node.source || !node.source.value) {
+					return;
+				} 
+				switch (node.source.value) {
+					case '/alloy':
+						this.imported.push('Alloy');
+						break;
+					case '/alloy/underscore':
+						this.imported.push('_');
+						break;
+					case '/alloy/backbone':
+						this.imported.push('Backbone');
+						break;
+				}
+			},
+			ReferencedIdentifier(path) {
+				const node = path.node;
+				if (toCheck.includes(node.name) && !this.required.includes(node.name) && !this.imported.includes(node.name)) {
+					this.required.push(node.name);
+					switch (node.name) {
+						case 'Alloy':
+							this.toRequire.push({
+								VARIABLE: 'Alloy',
+								REQUIRECALL: 'require(\'alloy\')'
+							});
+							break;
+						case '_':
+							this.toRequire.push({
+								VARIABLE: '_',
+								REQUIRECALL: 'require(\'alloy/underscore\')._'
+							});
+							break;
+						case 'Backbone':
+							this.toRequire.push({
+								VARIABLE: 'Backbone',
+								REQUIRECALL: 'require(\'alloy/backbone\')'
+							});
+							break;
+					}
+					
+				}
+			},
+			Program: {
+				enter() {
+					this.imported = [];
+					this.toRequire = [];
+					this.required = [];
+				},
+
+				exit(path) {
+					if (this.toRequire.length) {
+						for (const data of this.toRequire) {
+							path.unshiftContainer('body', [
+								buildRequire(data)
+							]);
+						}
+					}
+				}
+			}
+		}
+	};
+};
--- a/Alloy/commands/compile/ast/handle-alloy-globals.js
+++ b/Alloy/commands/compile/ast/handle-alloy-globals.js
@@ -53,19 +53,19 @@ module.exports = function(babel) {
 						case 'Alloy':
 							this.toRequire.push({
 								VARIABLE: 'Alloy',
-								REQUIRECALL: 'require(\'alloy\')'
+								REQUIRECALL: 'require(\'/alloy\')'
 							});
 							break;
 						case '_':
 							this.toRequire.push({
 								VARIABLE: '_',
-								REQUIRECALL: 'require(\'alloy/underscore\')._'
+								REQUIRECALL: 'require(\'/alloy/underscore\')._'
 							});
 							break;
 						case 'Backbone':
 							this.toRequire.push({
 								VARIABLE: 'Backbone',
-								REQUIRECALL: 'require(\'alloy/backbone\')'
+								REQUIRECALL: 'require(\'/alloy/backbone\')'
 							});
 							break;
 					}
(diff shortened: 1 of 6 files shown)


jquick-axway:
@ewanharris, Alloy also does code generation for each JS file under the "controllers" directory. It'll create a new version of those JS files where it injects `Alloy`, `Backbone`, and `_` as local variables at the top o the generated file. Do a build of an Alloy project and then look under the generated "Resources" directory and you'll see what I mean.

So, since we're explicitly adding these as globals now, perhaps we should remove those locals too?

sgtcoolguy:
@ewanharris my only concern here is backwards compatibility when there is no `global` object exposed. I believe we added it in the 7.0 time period? It'd be safer to do an if check on global if we want to ensure we work on older SDKs.

sgtcoolguy:
Relates to  https://jira.appcelerator.org/browse/TIMOB-26403 and appcelerator/titanium_mobile#10364

sgtcoolguy:
Related to the concern for backwards compatibility, global was added to iOs/Android in 7.1.0:
- https://github.com/appcelerator/titanium_mobile/pull/9689
- https://github.com/appcelerator/titanium_mobile/pull/9552

It was in Windows SDK long ago.

So I do think it's a valid concern that we actually do a check that `global` is defined before trying to hang these properties.

ewanharris:
@sgtcoolguy 5fb5d8ad794c73ca921fe28c3cc77db39ed4a8d3 adds a check for the existence of global before assigning to it

97a7c0f4ca28231ec041664de767556dc5c5d6e2 takes a slightly different route, during the babel pass over the files, it looks for usage of _, Alloy, and Backbone, requiring them if necessary (i.e. they are used, but the user hasn't required them). It doesn't appear to have much of an effect on the compile time of an app. 

What do you think of that, happy to back-out the commit if it's too wild 🙂

Alloy/commands/compile/ast/handle-alloy-globals.js
```diff
@@ -0,0 +1,94 @@
...
+			},
+			ReferencedIdentifier(path) {
+				const node = path.node;
+				if (toCheck.includes(node.name) && !this.required.includes(node.name) && !this.imported.includes(node.name)) {
+					this.required.push(node.name);
+					switch (node.name) {
+						case 'Alloy':
+							this.toRequire.push({
+								VARIABLE: 'Alloy',
+								REQUIRECALL: 'require(\'alloy\')'
```
sgtcoolguy: The require calls should use an absolute path, not the legacy variant here.

ewanharris: Updated (finally) 😀

sgtcoolguy:
wow, an oldie but goodie. @ewanharris I'm good with merging this if the conflict is resolved and it passes the tests...

Alloy/commands/compile/ast/handle-alloy-globals.js
```diff
@@ -0,0 +1,94 @@
...
+						case 'Alloy':
+							this.toRequire.push({
+								VARIABLE: 'Alloy',
+								REQUIRECALL: 'require(\'/alloy\')'
+							});
+							break;
+						case '_':
+							this.toRequire.push({
+								VARIABLE: '_',
+								REQUIRECALL: 'require(\'/alloy/underscore\')._'
```
brentonhouse: Why is there a '._' after the require call here?

ewanharris: underscore exports itself both as the main export [and as _](https://github.com/appcelerator/alloy/blob/515e9ed1160037749eceb9525b2cc2d78bd2d4e4/Alloy/lib/alloy/underscore.js#L52)

brentonhouse: ok.  Is there any advantage to doing this over just using the main export?

ewanharris:
@sgtcoolguy I'm thinking to maybe spend some time to see if we can remove the global assignment I initially added and get by with just the babel plugin to add the requires to files. What do you think?

Alloy/commands/compile/ast/handle-alloy-globals.js
ewanharris: No advantage AFAIK, just was what was familiar to me when I wrote the code

ewanharris:
@brentonhouse, I just pushed a commit that checks if the binding exists in the scope before we add the require. This should avoid the issue you're seeing where you're importing lodash as `_` and we assume you want to use `/alloy/underscore`.

Wanna try it out?

brentonhouse:
@ewanharris - Looks good!  Thanks!!

Alloy/commands/compile/ast/handle-alloy-globals.js
```diff
@@ -0,0 +1,98 @@
+const toCheck = [ 'Alloy', '_', 'Backbone' ];
+var template = require('@babel/template').default;
```
sgtcoolguy: If you're already using `const` above, why not also use `const` or `let` in place of these `var`s?

Alloy/commands/compile/ast/handle-alloy-globals.js
```diff
@@ -0,0 +1,98 @@
...
+			CallExpression: function (path, state) {
+				const node = path.node;
+				if (node.callee.name !== 'require') {
+					return;
+				}
+				if (!node.arguments || !node.arguments[0]) {
+					return;
+				}
+				switch (node.arguments[0].value) {
+					case '/alloy':
```
sgtcoolguy: Don't a lot of users have legacy/deprecated require calls with no leading '/'? May need to add extra cases without the leading '/'.

Alloy/commands/compile/ast/handle-alloy-globals.js
```diff
@@ -0,0 +1,98 @@
...
+						this.imported.push('Backbone');
+						break;
+				}
+			},
+			ImportDeclaration (path) {
+				const node = path.node;
+				if (!node.source || !node.source.value) {
+					return;
+				} 
+				switch (node.source.value) {
```
sgtcoolguy: extract the common code between import/require handling to a function that takes the raw string value?

Alloy/commands/compile/ast/handle-alloy-globals.js
```diff
@@ -0,0 +1,98 @@
+const toCheck = [ 'Alloy', '_', 'Backbone' ];
```
sgtcoolguy: This variable name is ok, but may make more sense with a name like `ALLOY_GLOBALS` or the wordier `GLOBALS_TO_CHECK`? super nit-picky style thing.

Alloy/commands/compile/ast/handle-alloy-globals.js
```diff
@@ -0,0 +1,98 @@
...
+								REQUIRECALL: 'require(\'/alloy/backbone\')'
+							});
+							break;
+					}
+					
+				}
+			},
+			Program: {
+				enter() {
+					this.imported = [];
```
sgtcoolguy: What the significance of the naming between `this.imported` and `this.required`? Don't both just keep track of explicit or implicit usage of the globals? Any reason to keep them separate?

Alloy/commands/compile/ast/handle-alloy-globals.js
```diff
@@ -0,0 +1,98 @@
...
+					case '/alloy/underscore':
+						this.imported.push('_');
+						break;
+					case '/alloy/backbone':
+						this.imported.push('Backbone');
+						break;
+				}
+			},
+			ReferencedIdentifier(path) {
+				const node = path.node;
```
sgtcoolguy: instead assign `const name = path.node.name;` and refer to that, rather than hold the node and keep asking for it's name property in all the references.

Alloy/commands/compile/ast/handle-alloy-globals.js
ewanharris: Agreed, went for a combination of the two with `ALLOY_GLOBALS_TO_CHECK` :)

Alloy/commands/compile/ast/handle-alloy-globals.js
ewanharris: Done, though about using object destructing here but didn't want to play around with the eslint config. (which reminds me to get round to finally finishing moving alloy to eslint-config-axway soon)

PR #913, opened 2018-10-02
[TIMOB-26403] Explicitly expose Alloy, _, and Backbone off the global object when it's available

sgtcoolguy:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-26403

**Description:**
This modifies the `app.js` template so that we don't simply declare `Alloy`, `_`, and `Backbone` as top-level variables, we also explicitly assign them as properties off `global` when there is a `global`.

We intend to remove exposing top-level app.js variables as implicit globals in the future (SDK 9?).

### DIFF
--- a/Alloy/template/app.js
+++ b/Alloy/template/app.js
@@ -1,11 +1,18 @@
 /**
  * Alloy for Titanium by Appcelerator
  * This is generated code, DO NOT MODIFY - changes will be lost!
- * Copyright (c) 2012 by Appcelerator, Inc.
+ * Copyright (c) 2012-2018 by Appcelerator, Inc.
  */
-var Alloy = require('/alloy'),
+var Alloy = require('/alloy');
 	_ = Alloy._,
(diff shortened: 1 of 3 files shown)


Alloy/template/app.js
```diff
@@ -1,11 +1,18 @@
 /**
  * Alloy for Titanium by Appcelerator
  * This is generated code, DO NOT MODIFY - changes will be lost!
- * Copyright (c) 2012 by Appcelerator, Inc.
+ * Copyright (c) 2012-2018 by Appcelerator, Inc.
  */
-var Alloy = require('/alloy'),
+var Alloy = require('/alloy');
 	_ = Alloy._,
 	Backbone = Alloy.Backbone;
```
feons: Why declare `_` and `Backbone` without `var`?

sgtcoolguy:
Closed in favor of #911

PR #901, opened 2018-08-21
[ALOY-1621] Move NavigationWindow API from Ti.UI.iOS to Ti.UI (parity)

feons:
https://jira.appcelerator.org/browse/ALOY-1621

PR #914, opened 2018-11-01
[ALOY-1621] Fix update min version and not die on non-ios platforms

feons:
https://jira.appcelerator.org/browse/ALOY-1621
https://jira.appcelerator.org/browse/ALOY-1636

ref: https://github.com/appcelerator/alloy/pull/901


jasonkneen:
@feons -- great -- fixes the issue now of using custom tags overrides for NavigationWindow on preview SDKs. thanks.

PR #915, opened 2018-11-13
[ALOY-1637] ES6 code frame fails when using duplicate variable declar…

feons:
…ations

https://jira.appcelerator.org/browse/ALOY-1637


ewanharris:
Merging, will bump version in separate PR

PR #922, opened 2019-03-07
[ALOY-1646] Add support for platform specific namespaces for xml attributes

brentonhouse:


Alloy/commands/compile/parsers/default.js
```diff
@@ -53,6 +53,17 @@ function parse(node, state, args) {
 		args.symbol = CU.generateUniqueId();
 	}
 
+	// find platform specific attributes
+	var platformAttributes = _.filter(_.keys(args.createArgs), key => _.includes(key, ':'));
```
ewanharris: Rather than looping through all the keys again in this file, I think it would be better to do this in the initial pass in [compilerUtils.getParserArgs](https://github.com/appcelerator/alloy/blob/784c0a7ea45510953533b1748dae0407316434b8/Alloy/commands/compile/compilerUtils.js#L116). Filtering out any args that aren't for the platform and removing the platform prefix from the attribute.

Alloy/commands/compile/compilerUtils.js
```diff
@@ -212,6 +212,18 @@ exports.getParserArgs = function(node, state, opts) {
 			});
 		} else {
 			var theValue = node.getAttribute(attrName);
+
+			// find platform specific attributes
+			var attributeParts = attrName.split(':');
+			if ( attributeParts.length === 2 ) {
```
ewanharris: Similar to line 164, we nee to check if the suspected platform `attributeParts[0]` is in `CONST.PLATFORMS`. Otherwise we are ignoring setting properties on `ListItems` combined with an `ItemTemplate`


```suggestion
			if ( attributeParts.length === 2 && _.includes(CONST.PLATFORMS, attributeParts[0])) {
```

PR #923, opened 2019-03-07
[ALOY-1647] Add support for camelCase, snake_case, and kabab-case XML elements

brentonhouse:


feons:
@brentonhouse, what benefits do this change bring?

This seems to introduce a hint of confusion especially to new developers. Also another opportunity for folks to argue over which case should be used.  Not to mention mixed usage in a file is not very easy on the eyes.

brentonhouse:
@feons - I would definitely want to encourage a best practice of consistency.   (i.e. Sticking with one case per file: PascalCase, camelCase, snake_case, or kabab-case.)

ewanharris:
I agree with Feon's points here, and would also like to add that something like this also negatively impacts the Atom/VS Code plugins and Studio. If we take this we'd have to update all the autosuggestions in those to handle suggesting these variants in view files (as well as the handling of detecting which to suggest, allowing setting a preferred choice incase etc).

cb1kenobi:
This is a really bad idea. We have a convention. If you use Alloy, you do PascalCase. Period.

With the Vue.js support, you use kebab-case. Period.

Supporting different conventions is confusing and antipattern. We will pass on this PR.

PR #924, opened 2019-03-13
fix(compile): do not normalize filepaths to be correct for platform in optimizeCompiledCode

ewanharris:
https://jira.appcelerator.org/browse/ALOY-1650

We use a regex to check exclude certain directories from being optimized here, that regex is constructed using posix path separators so by normalizing to windows separators we are no longer matching the regex

Fixes ALOY-1650


Sajoha:
LGTM... Don't have write access to merge though

PR #927, opened 2019-03-27
feat: support incremental builds

janvennemann:
This is a first approach to add incremental build capabilities to Alloy. I only took care of a part of the MVC builds step, specifically controller/view generation as those are usually the most files.

Other parts of the build pipeline could be handled in a similar manner. They just need to be split up in individual tasks and their logic needs to be adjusted to handle full/incremental builds.

Things that are not yet handled properly and need some more work:

- [ ] Handle deleted files and clear associated resources
- [ ] Force full builds if required when things other than input files change, like the Alloy compiler configuration for example.
- [ ] It may be possible to also speed up clean builds by doing processing in parallel instead of the current serial pipeline. This needs to be evaluated if the Alloy build process can handle that because it's completely serially right now.


hansemannn:
Any update on this? It's a real win on making larger projects more agile and scalable

ewanharris:
@hansemannn it wont make it into the next release (1.14.0) but maybe the one after that. There's still some work (as outlined in the PR) and a lot of testing to be done to ensure this doesn't cause any regressions.

hansemannn:
I'd be happy to test it with our own set of (larger) apps and cherry-picking it before the release would also not be an issue. Let me know if I can help.

hansemannn:
I wonder if a simple fill watcher on app/ couldn't help already. If a file changed, recompile that one, if not, skip.

The manual alloy compilation takes so much time for every run, it devastating.

PR #930, opened 2019-05-15
fix(ios): fix fetch() on iOS due to unbound call to ResultSet.field

sgtcoolguy:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-27053

**Description**
On Windows and Android we explicitly bound a reference to the ResultSet's field method to the instance
We did not on iOS, so changes to be more compliant on master resulted in an error
due to `this` being implicitly the global and not the ResultSet instance.
This is effectively the same issue as TIMOB-20577
Eliminated the redirection used to call the method altogether in favor of the more straight-forward
`rs.field(i)`.

Fixes TIMOB-27053


sgtcoolguy:
cc @hansemannn

PR #935, opened 2019-07-23
[ALOY-1698] Compile: Errors thrown during sourcemap step are not visible to a user

Sajoha:
https://jira.appcelerator.org/browse/ALOY-1698

### DIFF
--- a/Alloy/commands/compile/sourceMapper.js
+++ b/Alloy/commands/compile/sourceMapper.js
@@ -109,8 +109,7 @@ exports.generateCodeAndSourceMap = function(generator, compileConfig) {
 			sourceType: 'module'
 		});
 	} catch (e) {
-		logger.trace(genMap.code);
-		throw e;
+		U.dieWithCodeFrame(`Error parsing code in ${outfile}. ${e.message}`, e.loc, genMap.code);
 	}
 
 	// create source map and generated code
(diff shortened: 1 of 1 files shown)


Alloy/commands/compile/sourceMapper.js
```diff
@@ -208,8 +208,7 @@ exports.generateSourceMap = function(generator, compileConfig) {
 			sourceType: 'module'
 		});
 	} catch (e) {
-		logger.trace(genMap.code);
-		throw e;
+		U.dieWithCodeFrame('Error parsing code in "' + origFile.filename + '". Unexpected token at line ' + e.loc.line + ' column ' + e.loc.column, e.loc, genMap.code);
```
ewanharris: I think the Unexpected token at line was a bad move when I added this to the controller parsing code initially. Maybe we could make use of the `message` property on the error thrown from babel here? I think it should give better context to a user as to what the issue is. So something like

```js
U.dieWithCodeFrame(`Error parsing code in "${origFile.filename}". ${e.message}`, e.loc, genMap.code);
```

Sajoha:
So with the `generateCodeAndSourceMap` function an `origFile` property isn't passed. Instead I used the `target.filepath` value. Also switched to this in the `generateSourceMap` function for parity. However this value isn't as pretty, providing a full file path rather than an abbreviated one.

PR #938, opened 2019-09-24
fix(compile): allow return outside out function

brentonhouse:
Fix for part of issue in:  https://jira.appcelerator.org/browse/ALOY-1705


jasonkneen:
Whatever fixes this issue and allows us to use npm modules without this error is good for me.

PR #940, opened 2019-10-03
fix(compile): correctly handle platform prefix on event handlers

ewanharris:
Fixes [ALOY-1701](https://jira.appcelerator.org/browse/ALOY-1701)


brentonhouse:
Good catch!  Thanks @ewanharris !

PR #947, opened 2019-11-18
fix(compile): Define 'state' in ImportDeclaration

ssjsamir:
Ticket: https://jira.appcelerator.org/browse/ALOY-1710


ewanharris:
Note, the GA SDK test failed due to the Jenkins box using SDK 9, and us running tests under Node.js 8.x which is unsupported in that version. I'll update the Jenkinsfile accordingly

PR #949, opened 2019-12-18
fix: allow using widget as primary control in XML View

brentonhouse:


brentonhouse:
> This looks to have broken the use case [here](https://github.com/ewanharris/ALOY-1256-test).
> 
> Maybe we only want to do `code += '.getViewEx({recurse:true});\n';` when `type === 'widget'`, and maintain the existing behaviour for all other use cases?

This seems to be a greater issue of a require/widget code generator not being aware that it is generating code inside of a Ti.UI.Tab element.   Let me take a look and see if I can pass in some additional information from inside the `compile/parsers/Ti.UI.Tab.js` file.  That way the require/widget can have enough context to generate properly.

brentonhouse:
@ewanharris  - I've added support for when widgets/require are in Tabs now

PR #950, opened 2020-01-23
[TIMOB-26352]  Add support for using node_modules from project root directory

brentonhouse:


sgtcoolguy:
I don't think this is the right location to add support - it'd only work for Alloy apps. Also, it's leading to extra copying (copy from `node_modules` -> `Resources`, then "copy" from `Resources` -> app build dir)

I've started working on this in the SDK itself: appcelerator/titanium_mobile#11447

ewanharris:
Closing, as stated by @sgtcoolguy we'll be dealing with this at the SDK level

PR #951, opened 2020-01-23
[ALOY-1714] Allow ~/ to be used as a shortcut for WPATH() in xml view attributes

brentonhouse:
Allow `~/`  be used as a shortcut for this so that `WPATH()` is no longer required when using with XML View attributes

```xml
<Alloy>
    <ImageView id="image" image="~/images/a.png" />
</Alloy>
```


ewanharris:
@brentonhouse I think we're going to face similar challenges as #953 here. This currently only works for views, if we're adding this kind of syntactic sugar I think it should work in the majority of locations rather than just a single place

PR #952, opened 2020-01-24
refactor: use external compiler package

janvennemann:
WIP


Alloy/commands/compile/index.js
```diff
@@ -233,7 +239,6 @@ module.exports = function(args, program) {
 	if (restrictionPath === null) {
 		// Generate alloy.js from template
 		var libAlloyJsDest = path.join(paths.resources, titaniumFolder, 'alloy.js');
-		var pkginfo = require('pkginfo')(module, 'version');
```
ewanharris, line 236: As this was removed the version in the `alloy.js` file is currently blank, we could either add this back in or use `program.version()` to obtain the version from commander instead of `module.exports` (which is set using the same method as here ultimately)

ewanharris: Fixed in af691f2a2da00cdba67bd36b957b2857e458b66f

PR #953, opened 2020-01-24
[ALOY-1703]  Add several XML attribute aliases 

brentonhouse:


Alloy/commands/compile/parsers/Ti.UI.Button.js
```diff
@@ -36,6 +36,15 @@ function parse(node, state, args) {
 		node.setAttribute('systemButton', 'Ti.UI.' + iOSProxy + '.SystemButton.' + systemButton);
 	}
 
+	if ( node.hasAttribute('text') ) {
+		node.setAttribute('title', node.getAttribute('text'));
+		node.removeAttribute('text');
+	}
+
+	if ( node.getAttribute('verticalAlign') === 'center' ) {
+		node.setAttribute('verticalAlign', 'Titanium.UI.TEXT_VERTICAL_ALIGNMENT_CENTER');
```
ewanharris: Based on [Josh's comment](https://jira.appcelerator.org/browse/ALOY-1703?focusedCommentId=450997&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-450997), this is a change in behaviour correct?

Alloy/commands/compile/parsers/Ti.UI.Label.js
```diff
@@ -44,6 +44,10 @@ function parse(node, state, args) {
 		}
 	}
 
+	if ( node.getAttribute('verticalAlign') === 'center' ) {
+		node.setAttribute('verticalAlign', 'Titanium.UI.TEXT_VERTICAL_ALIGNMENT_CENTER');
```
ewanharris: Similar change in behaviour comment

Alloy/commands/compile/parsers/Ti.UI.TextArea.js
```diff
@@ -86,6 +86,10 @@ function parse(node, state, args) {
 			postCode += '<%= parentSymbol %>.add(' + parentSymbol + ');';
 		}
 
+		if ( node.getAttribute('verticalAlign') === 'center' ) {
+			node.setAttribute('verticalAlign', 'Titanium.UI.TEXT_VERTICAL_ALIGNMENT_CENTER');
```
ewanharris: Similar change in behaviour comment

Alloy/commands/compile/parsers/Ti.UI.TextField.js
```diff
@@ -136,6 +136,10 @@ function parse(node, state, args) {
 		}
 	}
 
+	if ( node.getAttribute('verticalAlign') === 'center' ) {
+		node.setAttribute('verticalAlign', 'Titanium.UI.TEXT_VERTICAL_ALIGNMENT_CENTER');
```
ewanharris: Similar change in behaviour comment

Alloy/commands/compile/parsers/Ti.UI.Button.js
brentonhouse: The thing is, I am not talking about changing anything with the SDK.  This is simply an intuitive way to be able to center text and it mirrors what you can do with textAlign.  I think NOT doing this is actually more harmful as most developers are going to assume that "center" can be used and won't find out until their build breaks.

Alloy/commands/compile/parsers/Ti.UI.Label.js
brentonhouse: > What's the desire for adding the `src` alias, is it purely because it's less characters than `image`? I'm not really a fan of adding an undocumented alias purely to avoid typing 2 characters.

I think src or source makes much more sense.  That would align us more with HTML, NativeScript, ReactNative, Xamarin, etc

ewanharris: If we want to use `src` that change should come in the SDK imo

Alloy/commands/compile/parsers/Ti.UI.Button.js
ewanharris: But there is a behaviour change here, `verticalAlign: 'center'` currently does one thing (which I agree doesn't do what I would expect) and here we're changing it do something completely different.

Alloy has no special handling for `textAlign`, I think it's that that `TEXT_ALIGNMENT_CENTER` has a value of `center` so it just works. `TEXT_VERTICAL_ALIGNMENT_CENTER` has a value (on Android) of `middle` so it doesn't work.

If we're adding special handling for `center`, why not add the other variants?

brentonhouse: The thing is, no one knows about "middle" unless you dig into the source code.  "center" makes sense because it is part of the name TEXT_VERTICAL_ALIGNMENT_CENTER and the fact that TEXT_ALIGNMENT_CENTER can be aliased by using "center".   If we want to change the value to TEXT_VERTICAL_ALIGNMENT_MIDDLE, then "middle" would make sense.

ewanharris: I wasn't saying people should know about middle, it was just an explanation of _why_ this weird behaviour happens.

I tried testing this change, and it looks like it only works when used in views and not in styles?

ewanharris:
@brentonhouse I like the idea of making this simpler. But I think as it stands, IMO the potential confusion from `verticalAlign: 'center'` not working everywhere (styles we could fix, js code we could try with a babel plugin but maybe at the cost of performance?) would outweigh the benefits

Topener:
To me it would be better to integrate these aliases into the SDK instead of Alloy. It is hard to explain in documentation some things exist in alloy only. I also think introducing things like a src property on an ImageView will cause more confusion than it solves. There already is confusion with html, we shouldn't want to confirm that.

brentonhouse:
understood.

PR #956, opened 2020-03-06
[ALOY-1722] Make alloy work as required module

brentonhouse:


janvennemann:
Now the Alloy require still has an indirect dependency on `commander` as the constructor expects a configured and parsed `program`. That doesn't feel very intuitive.

I guess you are just trying to run Alloy compile on a project? I'd say a simple `compile` function that accepts the project path and platform (and whatever else it may need) would be a more sophisticated API, don't you think. So something like this:

```js
const { compile } = require('alloy');
compile({
  project: '/path/to/my/app'
  platform: 'ios'
});
```

Also, we should think about what part's of the Alloy CLI actually make sense to expose as a proper node API. I know this is just a quick and dirty solution, but as you see there are more things to consider.

brentonhouse:
@janvennemann - correct.  I think the parameters and other things should be updated to that the CLI  takes its parameters and correctly passes those to the alloy file.  Unfortunately, when this was built, it was written to expect the "program" object from commander to be passed around to files all over in the library.  I'd say this is a good first step and can be merged now as it is better than the existing solution and doesn't break anything.

ewanharris:
I would prefer for us to deal with this as planned in Alloy 2 alongside #952. I don't think we should make the commander definition the interface here but instead as Jan suggested provide a simpler interface that abstracts that implementation detail (commander) away.

Rather than try and rush to provide a working solution on 1.x, let's work together (and communicate before we land a PR) to find something that fits the goals we all want from 2.x.

janvennemann:
In the meantime, if you really need to require Alloy commands right now, have you tried requiring the command you need directly? Wouldn't that have the same effect?

```js
const compile = require('alloy/Alloy/commands/compile');
const program = {};
compile(program);
```

You loose some sanity checks regarding banner/help display, but i assume you know what you are doing so you can make sure to pass a properly configured program to the command.

PR #957, opened 2020-03-20
Remove os:windows from config.json

Topener:
Removing os:windows from config.json due to dropping Windows support since SDK 9.0


ewanharris:
I think we should also completely remove the Windows platform in #952

PR #960, opened 2020-06-30
refactor: fix invalid version comparisons

ewanharris:
JIRA: [ALOY-1720](https://jira.appcelerator.org/browse/ALOY-1720)

* Matrix calls - As we were using the version to basically do feature detection I went with checking if the functions existed rather than using an SDK version check
* M2_320 check - As we've dropped support for Mobilweb _and_ 3.2.0 is years old, I went with removing the version check and just having it be a `OS_MOBILEWEB` check

### DIFF
--- a/Alloy/builtins/animation.js
+++ b/Alloy/builtins/animation.js
@@ -21,7 +21,7 @@ exports.HORIZONTAL = 'horizontal';
 exports.VERTICAL = 'vertical';
 
 const create3DMatrix = Ti.UI.createMatrix3D ? Ti.UI.createMatrix3D : Ti.UI.create3DMatrix;
-const create2DMatrix = Ti.UI.createMatrix3D ? Ti.UI.createMatrix2D : Ti.UI.create2DMatrix;
+const create2DMatrix = Ti.UI.createMatrix2D ? Ti.UI.createMatrix2D : Ti.UI.create2DMatrix;
 
 /**
  * @method flip
--- a/Alloy/builtins/animation.js
+++ b/Alloy/builtins/animation.js
@@ -20,8 +20,8 @@ exports.HORIZONTAL = 'horizontal';
  */
 exports.VERTICAL = 'vertical';
 
-const create3DMatrix = Ti.version >= '8.0.0' ? Ti.UI.createMatrix3D : Ti.UI.create3DMatrix;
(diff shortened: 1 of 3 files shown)


Alloy/template/lib/alloy.js
```diff
@@ -32,9 +32,8 @@ exports._ = _;
 exports.Backbone = Backbone;
 
 var DEFAULT_WIDGET = 'widget';
-var TI_VERSION = Ti.version;
-var MW320_CHECK = OS_MOBILEWEB && TI_VERSION >= '3.2.0';
-var IDENTITY_TRANSFORM = OS_ANDROID ? (Ti.version >= '8.0.0' ? Ti.UI.createMatrix2D() : Ti.UI.create2DMatrix()) : undefined;
+var MW320_CHECK = OS_MOBILEWEB;
+var IDENTITY_TRANSFORM = OS_ANDROID ? (Ti.UI.createMatrix3D ? Ti.UI.createMatrix2D() : Ti.UI.create2DMatrix()) : undefined;
```
jquick-axway, line 36: You should be checking for `Ti.UI.createMatrix2D` existence.
I think this is a copy-and-paste mistake. :stuck_out_tongue:

Alloy/builtins/animation.js
```diff
@@ -20,8 +20,8 @@ exports.HORIZONTAL = 'horizontal';
  */
 exports.VERTICAL = 'vertical';
 
-const create3DMatrix = Ti.version >= '8.0.0' ? Ti.UI.createMatrix3D : Ti.UI.create3DMatrix;
-const create2DMatrix = Ti.version >= '8.0.0' ? Ti.UI.createMatrix2D : Ti.UI.create2DMatrix;
+const create3DMatrix = Ti.UI.createMatrix3D ? Ti.UI.createMatrix3D : Ti.UI.create3DMatrix;
+const create2DMatrix = Ti.UI.createMatrix3D ? Ti.UI.createMatrix2D : Ti.UI.create2DMatrix;
```
jquick-axway, line 24: There's a typo in the 2nd line. You should be checking for the 2D method's existence, not the 3D method.

ewanharris: Doh, fixed!

ewanharris:
Merging as needs to be released for QE testing

PR #963, opened 2020-08-04
Add templates for webpack

ewanharris:
JIRA: https://jira.appcelerator.org/browse/ALOY-1733

* Updates the `alloy new` command to have special handling for webpack templates
	* Setup package.json if one doesn't exist
	* Add the required dependencies to the project
	* Install dependencies
	* Add new eslint config if it exists in the template directory
* Add new template `webpack-default` which is an alloy port of the classic on created in appcelerator/titanium_mobile/pull/11858\
	* I duplicated the assets, as there's no guarantees the user actually created the right classic template
* Was forced to update eslint also as the previous version we were on did not know async/await was a thing
* Add the template to the `alloy info template` output so that it shows in Studio

Test steps:

Ti CLI:
	
	1. Create the webpack template `ti create --type app --template webpack-default`
	2. cd into the project
	3. `alloy new . --template webpack-default`

Appc CLI (once merged in):

	1. `appc new --type titanium --template webpack-default`

Once approved, I'll push a commit to update the versions and then update appc cli

Appc Studio (once merged in):

	1. Create a new alloy app and select the `Alloy Project with webpack` template
	2. Fill out details
	3. Wait


templates/webpack-default/app/controllers/index.js
```diff
@@ -0,0 +1,5 @@
+const tabgroup = require('@/tabgroup');
+
+tabgroup.tg = $.index;
```
Topener, line 3: Could change this without assigning a variables and setting it straight to the source 
```js
require('@/tabgroup').tg = $.index
```

templates/webpack-default/app/views/home/index.xml
```diff
@@ -0,0 +1,14 @@
+<Alloy>
+	<Tab title="Home">
+		<Window title="Home">
+			<View id="wrapper">
+				<Label class="fa-solid search">edit</Label>
+				<Label id="editLabel">
+					<AttributedString class="as">
+						Edit home.js to change this view!
```
janvennemann, line 8: This should be `views/home/index.xml`. Although technically there are three different files that make up the view, but that probably blows up the example.

templates/webpack-default/app/views/profile/index.xml
```diff
@@ -0,0 +1,31 @@
...
+				<ImageView id="profile" />
+				<Label class="profileName">Verona Blair</Label>
+				<TableView onClick="doClick">
+					<TableViewSection id="settingsSection">
+						<HeaderView>
+							<View class="separatorWrapper">
+								<View class="separator" />
+							</View>
+						</HeaderView>
+						<Require src="/profile/row" iconBG="#eef0f6" icon="user" iconColor="#a9cdd1" label="Personal Details" labelColor="#000" />
```
janvennemann, line 14: I know we don't have any style rules for this, but the prettier setting for Vue for example tends to wrap attributes to new lines ones the line gets over a certain amount of character. I find that easier to read than a single line that goes on and on.

```suggestion
						<Require
							src="/profile/row"
							iconBG="#eef0f6"
							icon="user"
							iconColor="#a9cdd1"
							label="Personal Details"
							labelColor="#000"
						/>
```

templates/webpack-default/app/README
```diff
@@ -0,0 +1,31 @@
+Welcome to Alloy. Prepare to be amazed.
```
janvennemann, line 1: The default webpack template contains a `README.md` so having an additional `README` is probably confusing. I aligned the `README.md` for the new Angular and webpack classic templates so they are bascially the same, just point to different guides where to get help. Would be nice if the Alloy readme looks the same.

templates/webpack-default/app/views/home/index.xml
ewanharris: Doi of course!

templates/webpack-default/app/controllers/index.js
ewanharris: I've updated this file as it's a single usage, but kept the one in profile/index.js as it's used in the click handler

templates/webpack-default/app/views/profile/index.xml
ewanharris: Updated the Require tags to look like that

templates/webpack-default/app/README
ewanharris: Updated the `alloy new` command to append to/create the `README.md` file if one exists in the app template - 7d82a5c67c7ecbf6fe0b0dd3c6ee264257356e09

PR #964, opened 2020-08-05
fix(runtime): use bootstrap mechanism to load globals rather than setting in the alloy.js file

ewanharris:
Fixes [ALOY-1734](https://jira.appcelerator.org/browse/ALOY-1734)

* `alloy.bootstrap.js` will be loaded at startup before the `app.js`, this ensures that the globals are correctly configured _before_ the `app.js` is even executed.
* Check if the globals exist in the `app.js` and set them if they don't, this should keep us compatible with SDKs pre-7.5.0 (i.e. those without the bootstrap feature, if you're using an SDK that old ... please, please, please update ASAP)

Test steps - see jira

### DIFF
--- a/Alloy/commands/compile/index.js
+++ b/Alloy/commands/compile/index.js
@@ -480,6 +480,8 @@ module.exports = function(args, program) {
 
 	generateAppJs(paths, compileConfig, restrictionPath, compilerMakeFile);
 
+	U.copyFileSync(path.join(alloyRoot, 'template', 'alloy.bootstrap.js'), path.join(paths.resources, titaniumFolder, 'alloy.bootstrap.js'));
+
 	// ALOY-905: workaround TiSDK < 3.2.0 iOS device build bug where it can't reference app.js
 	// in platform-specific folders, so we just copy the platform-specific one to
 	// the Resources folder.
--- a/Alloy/template/alloy.bootstrap.js
+++ b/Alloy/template/alloy.bootstrap.js
@@ -1,0 +1,5 @@
+var Alloy = require('/alloy');
+
+global.Alloy = Alloy;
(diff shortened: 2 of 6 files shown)


brentonhouse:
@ewanharris @jquick-axway  - I know you have already merged this but I've tried using bootstrap files for assigning globals in the past and although it works when running normally, it breaks everything when running in LiveView.

jquick-axway:
@brentonhouse, I'm not sure how it could fail. When you make a file change, LiveView invokes Titanium's undocumented `Ti.App._restart()` method, which then reloads our "ti.main.js" script, which in turn loads all of our bootstrap scripts before requiring-in the "app.js". I've tested Android with and hyperloop (which now uses a bootstrap script and binding APIs) and it restarts with LiveView fine for me.

But it doesn't hurt to re-test Alloy and LiveView again anyways. It's part of our smoke test anyways.

brentonhouse:
What happens is there becomes multiple instances of modules.  If you set a property on a module in alloy.js and go to read it in index.js, the property is not set as it is a clean instance of the module.

brentonhouse:
What happens is there becomes multiple instances of modules.  If you set a property on a module in alloy.js and go to read it in index.js, the property is not set as it is a clean instance of the module.

brentonhouse:
We saw this before but I think it was fixed by this https://github.com/appcelerator/alloy/pull/955

ewanharris:
@brentonhouse I'm unaware of any issues between alloy and the boostrap mechanism, I know we've seen issues similar to what you're describing ([TIMOB-27206](https://jira.appcelerator.org/browse/TIMOB-27206)), but that was specifically down to liveview and it's inability to handle node_modules.

Do you have

PR #965, opened 2020-08-07
Ensure package.json has required stanzas, create vscode settings dir on project creation

ewanharris:
* Make sure the fields we try to access exist before accessing them - [ALOY-1736](https://jira.appcelerator.org/browse/ALOY-1736)
* Create `.vscode/settings.json` when running alloy new - [ALOY-1735](https://jira.appcelerator.org/browse/ALOY-1735)

@jquick-axway, could you check that the `settings.json` file is correct?


Alloy/commands/new/index.js
```diff
@@ -209,6 +227,14 @@ module.exports = async function(args, program) {
 		await fs.copy(path.join(paths.eslintTemplate), path.join(paths.eslintApp));
 	}
 
+	// Copy across the settings.json to have VS Code hide certain directories
+	const vscodeDir = path.join(paths.project, '.vscode');
+	if (!await fs.exists(vscodeDir)) {
```
brentonhouse, line 232: I am not sure you want to make this mandatory.  I NEVER store vscode settings locally and I wouldn't ever want to exclude the build folder from my file explorer as I used it all the time.

jquick-axway: @brentonhouse , a common mistake I see 1st time Alloy users make is they make changes to files under the root `i18n` and `platform` folders, only to be confused as to why their changes didn't work. Hiding these folders by default will hopefully encourage devs to make these changes under the `app` folder instead.

Also, we've heard of file locking issues happening on Windows under the generated folders which can cause build failures. We believe it's because of VS Code's file-monitoring/indexing. I'm not sure if this still happens today, but from what we've seen by other VS Code users in the past (not Titanium users) this may likely be the case and disabling indexing of these folders (namely `build`) works-around it.

If you don't want the `.vscode/settings.json` in your new project, then delete the file. Most people will make major changes to the created app project afterwards anyways.

jquick-axway:
@ewanharris, I've modified the Titanium SDK's PR to exclude the `node_modules` folder and `package-lock.json` file.
https://github.com/appcelerator/titanium_mobile/pull/11870/commits/bc8797112666b4ffc0ad50df4f8f4154ee7f5261

Alloy/commands/new/index.js
ewanharris: This is already occuring in Studio and I don't think I've ever seen anyone complain about Studio hiding those folders. Nonetheless, I've added a note to the default readme to let people know that it's there and that they can delete it if they wish

PR #966, opened 2020-08-11
fix: Use items in ListSection 

m1ga:
JIRA: https://jira.appcelerator.org/browse/AC-6585

Use `items = ` instead of deprecated `setItems()`. Will get rid of the warnings.


m1ga:
:+1: at least in my app I'm not seeing any other warning but it is always good to check, not using everything of course :smile:

PR #967, opened 2020-08-28
fix: handle widget case explicitly and revert back to previous default behaviour

ewanharris:
Fixes [ALOY-1738](https://jira.appcelerator.org/browse/ALOY-1738)

cc @brentonhouse I've verified that [brentonhouse/alloy-widget-test](https://github.com/brentonhouse/alloy-widget-test) still work with this change, do you have any other usage of top levels widgets?


ewanharris:
Marking as draft as the usage of the [infinite scroll widget](https://github.com/Topener/nl.fokkezb.infiniteScroll/tree/test) is not fixed by this PR

PR #968, opened 2020-09-30
chore(new): remove splash screens from template

Topener:
JIRA: https://jira.appcelerator.org/browse/ALOY-1741
A few changes have been made, including stripping out all iOS splash screens as Apple will reject any app using those anyways. Furthermore removed mobileweb and Windows and added readme's to both iOS and Android assets directory to help people along the way.


ewanharris (approved):
Changes look good to me! We should probably do something similar in the SDK to remove theses completely

PR #1035, opened 2020-11-18
fix: improve check around whether widget is only tag in the view

ewanharris:
We now look for whether there are any sibling nodes alongside the Widget tag, if there are then
we should not make this Widget the top level view

Fixes ALOY-1743

### DIFF
--- a/Alloy/commands/compile/parsers/Alloy.Require.js
+++ b/Alloy/commands/compile/parsers/Alloy.Require.js
@@ -134,7 +134,7 @@ function parse(node, state, args) {
 	};
 	if (args.parent.symbol && !state.templateObject && !state.androidMenu) {
 		code += ';\n' + args.symbol + '.setParent(' + args.parent.symbol + ');\n';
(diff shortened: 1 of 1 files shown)


Alloy/commands/compile/parsers/Alloy.Require.js
```diff
@@ -148,3 +148,36 @@ function parse(node, state, args) {
 		code: code
 	};
 }
+
+/**
+ * Determines whether a node is the only child node of the Alloy in the view structure. This is
+ * used to determine whether a Widget should become the root element of the view when only a view
+ * only contains a single Widget tag like below
```
janvennemann, line 155:
```suggestion
 * Determines whether a node is the only child node of the Alloy tag in the view structure. This is
 * used to determine whether a Widget should become the root element of the view when a view
 * only contains a single Widget tag like below
```

janvennemann (approved):
Super nit-picky note on the comment. The actual code changes are looking good!

PR #1072, opened 2021-01-12
Fix for Ti >=9; Backbone no longer global

dbankier:
As of Ti 9, globals are not supported. Without this fix, the properties adapter no longer works as Backbone would be undefined.

### DIFF
--- a/Alloy/lib/alloy/sync/properties.js
+++ b/Alloy/lib/alloy/sync/properties.js
@@ -1,6 +1,7 @@
 var Alloy = require('/alloy'),
 	_ = require('/alloy/underscore')._,
-	TAP = Ti.App.Properties;
+	TAP = Ti.App.Properties,
+  Backbone = Alloy.Backbone;
 
 function S4() {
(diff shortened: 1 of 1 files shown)


ewanharris:
@dbankier, I'm unable to reproduce this. Backbone should always be assigned to the globals either [here](https://github.com/appcelerator/alloy/blob/master/Alloy/template/alloy.bootstrap.js#L5) or [here](https://github.com/appcelerator/alloy/blob/master/Alloy/template/app.js#L15) dependent on SDK version.

I'm testing using [this sample](https://github.com/appcelerator-developer-relations/appc-sample-databinding) and adding code to call the sync method

dbankier:
Try using properties adapter. You will get Backbone is undefined.

ewanharris:
@dbankier, that sample [does](https://github.com/appcelerator-developer-relations/appc-sample-databinding/blob/master/app/models/Fruit.js#L5), and I added logging to make doubly sure 🙂 

We shipped a fix for Alloy to set the globals correctly in #964

dbankier:
Ok. Sorry then. Must be using an older version of alloy. Closing. Thanks.

PR #1218, opened 2021-09-06
fix(compile): provide filename to babel transform to allow using config files

ewanharris:
Fixes ALOY-1749

Test steps in jira

### DIFF
--- a/Alloy/commands/compile/sourceMapper.js
+++ b/Alloy/commands/compile/sourceMapper.js
@@ -63,6 +63,17 @@ exports.generateCodeAndSourceMap = function(generator, compileConfig) {
 		file: `${compileConfig.dir.project}/${relativeOutfile}`,
 		sourceRoot: compileConfig.dir.project
 	});
+	// try to lookup the filename, falling back to the output file if we can't determine it
+	let filename;
+	if (data.__MAPMARKER_CONTROLLER_CODE__) {
+		filename = data.__MAPMARKER_CONTROLLER_CODE__.filename;
+	} else if (data.__MAPMARKER_ALLOY_JS__) {
+		filename = data.__MAPMARKER_ALLOY_JS__.filename;
+	} else if (data.__MAPMARKER_NONCONTROLLER__) {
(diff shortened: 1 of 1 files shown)


Alloy/commands/compile/sourceMapper.js
```diff
@@ -63,6 +63,17 @@ exports.generateCodeAndSourceMap = function(generator, compileConfig) {
...
 		sourceRoot: compileConfig.dir.project
 	});
+	// try to lookup the filename, falling back to the output file if we can't determine it
+	let filename;
+	if (data.__MAPMARKER_CONTROLLER_CODE__) {
+		filename = data.__MAPMARKER_CONTROLLER_CODE__.filename;
+	} else if (data.__MAPMARKER_ALLOY_JS__) {
+		filename = data.__MAPMARKER_ALLOY_JS__.filename;
+	} else if (data.__MAPMARKER_NONCONTROLLER__) {
+		filename = data.__MAPMARKER_NONCONTROLLER__.filename;
```
garymathews, line 73: Shame we can't do optional chaining here since our min node version is 12, but we should check `filename` isn't `null`
```suggestion
	if (data.__MAPMARKER_CONTROLLER_CODE__ && data.__MAPMARKER_CONTROLLER_CODE__.filename) {
		filename = data.__MAPMARKER_CONTROLLER_CODE__.filename;
	} else if (data.__MAPMARKER_ALLOY_JS__ && data.__MAPMARKER_ALLOY_JS__.filename) {
		filename = data.__MAPMARKER_ALLOY_JS__.filename;
	} else if (data.__MAPMARKER_NONCONTROLLER__ && data.__MAPMARKER_NONCONTROLLER__.filename) {
		filename = data.__MAPMARKER_NONCONTROLLER__.filename;
```

garymathews (changes requested):

Alloy/commands/compile/sourceMapper.js
ewanharris: Updated!

garymathews (approved):
CR: **PASS**

PR #1243, opened 2021-11-01
feat(all): set default backbone version

m1ga:
This way new apps will use the newer Backbone version and not the very old default 0.9.2! Old/existing apps will still use 0.9.2 when no other version is defined and still work. Raising the constant in https://github.com/appcelerator/alloy/blob/master/Alloy/common/constants.js#L49 could break existing apps.

### DIFF
--- a/Alloy/template/config.json
+++ b/Alloy/template/config.json
@@ -5,5 +5,6 @@
 	"env:production": {},
 	"os:android": {},
 	"os:ios": {},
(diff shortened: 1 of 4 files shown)


ewanharris (approved):
This makes sense to me, I'm going to retarget the branch so it and #1237 can release together

m1ga:
Awesome! Thanks for the quick merge :+1:

PR #1326, opened 2022-03-29
remove ASCII art banner

caspahouzer:


hansemannn:
Why should we remove it? It's a cool piece of legacy!

cb1kenobi:
@hansemannn that ascii art is so 2013!

caspahouzer:
So... what we are doing?

cb1kenobi (approved):

PR #1333, opened 2022-04-12
ci: add github action to publish alloy

ewanharris:
This adds a publish action similar to the others that will publish the alloy package when a new release is created.

I've also added a `CONTRIBUTING.md` file that describes how to do this, it totally needs fleshing out some more but maybe that can be done after the release.

It also removes the `-b latest` usage from the test matrix cos it no longer exists.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
       matrix:
         nodeVersion: [ '12.x' ]
         os: [ macos-latest ]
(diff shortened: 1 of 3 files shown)


cb1kenobi:
I'm already working on a standard publish action that supersedes this method. I would prefer to use the new hotness instead of this.

ewanharris:
@cb1kenobi how far out is that from completion? If it's soon then sure

cb1kenobi (approved):
Looks good! **APPROVED**!

PR #1346, opened 2022-10-27
build: bump version

hansemannn:
@cb1kenobi The missing piece!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "alloy",
-  "version": "1.18.0",
+  "version": "2.0.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "alloy",
-  "version": "1.18.0",
+  "version": "2.0.2",
```
hansemannn, line 3: Not sure why this was still on 1.x, but now it should be clean!

cb1kenobi (approved):

package-lock.json
cb1kenobi: The package version is also in the package-lock, but when we manually bump the version in GitHub, we don't bump it in the lock file.

PR #1347, opened 2022-10-31
Package update

m1ga:
update node packages.

`npm run tests` is the same output with master and this PR:  `2825 specs, 28 failures`

**Changes:**

before:
```
18 vulnerabilities (1 low, 1 moderate, 10 high, 6 critical)
```

after:
```
4 vulnerabilities (1 low, 3 critical)
```


m1ga:
That's annoying :smile:  Code doesn't match because of spaces:

code in:
```
var a,
  b = "iPhone OS" == "iPhone OS",
  c = 2;
```

code out:
```
var a,
b = "iPhone OS" == "iPhone OS",
c = 2;
```

let's see if I can change that...

m1ga:
There we go `2361 specs, 0 failures` :+1: 
I'm doing a follow up PR where I remove old test folders for mobileweb and windows. But in order to keep this one clean I do it in a own PR

ewanharris (approved):

PR #1358, opened 2023-01-11
feat: update momentjs

m1ga:
* update moment.js from 2.24.0 to 2.29.4 ([changelog](https://github.com/moment/moment/blob/develop/CHANGELOG.md))
* update moment.js lang files (+ some new files)

Alloy tests ran fine:
```
Finished in 585.561 seconds
2361 specs, 0 failures
```

edit:
* increase test timeout for the actions


Alloy/builtins/moment/lang/af.js
```diff
@@ -1,72 +1,71 @@
...
+//! locale : Afrikaans [af]
+//! author : Werner Mollentze : https://github.com/wernerm
 
-;(function (global, factory) {
-   typeof exports === 'object' && typeof module !== 'undefined'
-       && typeof require === 'function' ? factory(require('/alloy/moment')) :
-   typeof define === 'function' && define.amd ? define(['../moment'], factory) :
-   factory(global.moment)
-}(this, (function (moment) { 'use strict';
+import moment from '../moment';
```
ewanharris, line 5: I'm not sure if it's worthwhile being concerned about the switch to ESM here, Android should work fine and I guess babel should transpile for iOS but I'd be curious to see how this might impact build times if we transpile

m1ga: Looks like the old version used the files from `locale` instead of `dist/local`. I've change that.

And in case we'll update them later again: there is one change inside the files:
`require('../moment')` is `require('/alloy/moment')`. Can be changed with search & replace.

ewanharris (approved):

PR #1360, opened 2023-03-04
feat(android): add collapseToolbar support

m1ga:
* Add `CollapseToolbar` to Alloy (copied over from `DrawerLayout`)

needed for https://github.com/tidev/titanium_mobile/pull/13760

```xml
<Alloy>
	<Window>

		<CollapseToolbar platform="android">

			<ContentView>
				<View backgroundColor="red" height="Ti.UI.SIZE">
					<Label>test</Label>
				</View>
			</ContentView>
		</CollapseToolbar>

	</Window>
</Alloy>
```

### DIFF
--- a/Alloy/commands/compile/parsers/Ti.UI.Android.CollapseToolbar._ProxyProperty.js
+++ b/Alloy/commands/compile/parsers/Ti.UI.Android.CollapseToolbar._ProxyProperty.js
@@ -1,4 +1,4 @@
-var _ = require('lodash');
+const _ = require('lodash');
 
 exports.parse = function(node, state) {
 	_.extend(state, {
--- a/Alloy/commands/compile/parsers/Ti.UI.Android.CollapseToolbar._ProxyProperty.js
+++ b/Alloy/commands/compile/parsers/Ti.UI.Android.CollapseToolbar._ProxyProperty.js
@@ -4,7 +4,7 @@ exports.parse = function(node, state) {
 	_.extend(state, {
 		proxyPropertyDefinition: {
 			parents: [
-				'Ti.UI.Android.DrawerLayout'
+				'Ti.UI.Android.CollapseToolbar'
 			]
 		}
 	});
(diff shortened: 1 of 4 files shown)


Alloy/commands/compile/parsers/Ti.UI.Android.CollapseToolbar._ProxyProperty.js
```diff
@@ -0,0 +1,12 @@
+var _ = require('lodash');
+
+exports.parse = function(node, state) {
+	_.extend(state, {
```
ewanharris, line 4: I think it'd be worthwhile guarding this against usage in unsupported SDKs, there's an example in [Ti.UI.OptionBar](https://github.com/tidev/alloy/blob/792ba0fa74e5a32d1c4ee2a6cce3ece30728c70d/Alloy/commands/compile/parsers/Ti.UI.OptionBar.js#L7-L10C3), I _think_ it'd look like as follows

```
	const MIN_VERSION = '12.1.0';
	const tiappSdkVersion = tiapp.getSdkVersion();
	if (tiapp.version.lt(tiappSdkVersion, MIN_VERSION)) {
		U.die(`Ti.UI.CollapseToolbar requires Titanium SDK ${MIN_VERSION}+`);
	}
```

m1ga:
my local tests are failing for the the (ios) popover test. It always says
```
[ERROR] Error with <ContentView> at line 3
[ERROR] Parent element must one of the following: [Ti.UI.Android.CollapseToolbar]
```
but I've added both in:

https://github.com/tidev/alloy/pull/1360/files#diff-2275f483800363596424aa8298ec1149ebb373c05b008848a397f859745e75a8R4-R8

ewanharris:
Hmm yeah it works for me too, are you running the tests on Linux? CI runs on mac and so do I so maybe there's something there?

m1ga:
Sorry for the late reply. 
Yes, I'm running the test on Linux. `PopOver` is iOS only and `CollapseToolbar` is Android only so the error won't happen in real apps anyway. Even though about adding `platform="ios"` to the xml to trick the test :smile: 

I'll put it on my todo list to fix the test. Just don't understand why it doesn't list both parents for me :shrug:

m1ga:
:facepalm: 

<b>fixed my tests!</b>
It used the installed parser files and not the changed one in the project. Had to do a `npm i -g alloy/` to have them in the right folder!

All green :+1:

ewanharris:
Ahh that explains why it worked for me as I tend to have alloy installed via `npm link`, I think I just realised a bug in the way the tests add the alloy executable in this project to the path so I'll push a fix for that!

ewanharris (approved):

PR #1361, opened 2023-03-11
test: prefix bin to path to ensure local alloy is used

ewanharris:
Executables are resolved first found in the path iirc, so prefixing is the right thing to do


m1ga (commented):
LGTM :+1: 

This way it will read my local parser files from the current Alloy folder without the need to link it first!

PR #1364, opened 2023-08-11
fix: missing includes

m1ga:
**Changes**
* add missing includes.

**Fixes**
Being able to use `<ContentView ns="Ti.UI.Android.CollapseToolbar">` in Alloy again.

---
**Sidenote**
Still have to find out why `isTitanium` is false and the `Titanium` namespace is not available in
https://github.com/tidev/alloy/blob/master/Alloy/common/constants.js#L238-L239
it used to work.

### DIFF
--- a/Alloy/commands/compile/parsers/Ti.UI.Android.CollapseToolbar.ContentView.js
+++ b/Alloy/commands/compile/parsers/Ti.UI.Android.CollapseToolbar.ContentView.js
@@ -1,3 +1,5 @@
+const _ = require('lodash');
+
 exports.parse = function(node, state) {
 	_.extend(state, {
 		proxyPropertyDefinition: {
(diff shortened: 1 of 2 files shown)


ewanharris:
@m1ga The `isTitanium` thing is a little weird, I think that's intended to be there for when the file is used at runtime in an app, the only other thing that uses `isTitanium` is a mobileweb check so it's basically unused.

I can't figure out a way to get that check working as when the file is first required we have no compile config to check which platform is being built for, I wonder if there's anyway that the file can be split to a build time and run time file that would potentially allow us to workaround not having the compile config?

m1ga:
I see, thanks for the info. I thought it was a general `isTitanium` check. I'll check if I can find another way to check for Android

ewanharris (approved):

PR #1365, opened 2023-10-10
Use SWC to parse and transform code instead of Babel

ewanharris:
This is a proof of concept in replacing Babel in the compilation step, it probably is not ready for merge given it's limited testing (hence the draft PR) but I think it's at a stage for potential discussion.

The positives here are mostly around the faster build time, I am limited on what I have to test but there appears to be a consistent speed up between the Babel based Alloy and SWC based Alloy (benchmarks and testing method below)

#### Default app

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `alloy compile /Users/awam/Documents/titanium-workspace/test-app/app --config platform=ios,version=0,simtype=none,devicefamily=universal,deploytype=development,target=simulator` | 280.0 ± 5.7 | 273.6 | 293.1 | 1.52 ± 0.03 |
| `alloy2 compile /Users/awam/Documents/titanium-workspace/test-app --config platform=ios,version=0,simtype=none,devicefamily=universal,deploytype=development,target=simulator` | 183.9 ± 1.8 | 181.6 | 187.0 | 1.00 |

<details>
<summary>Benchmark command</summary>
`hyperfine --prepare 'ti clean' --runs 10 'alloy compile /Users/awam/dev/git/tidev/kitchensink-v2/app --config platform=ios,version=0,simtype=none,devicefamily=universal,deploytype=development,target=simulator' 'alloy2 compile /Users/awam/dev/git/tidev/kitchensink-v2/app --config platform=ios,version=0,simtype=none,devicefamily=universal,deploytype=development,target=simulator'`
</details>

#### KitchenSink

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `alloy compile /Users/awam/dev/git/tidev/kitchensink-v2/app --config platform=ios,version=0,simtype=none,devicefamily=universal,deploytype=development,target=simulator` | 879.0 ± 18.7 | 859.2 | 918.6 | 1.52 ± 0.03 |
| `alloy2 compile /Users/awam/dev/git/tidev/kitchensink-v2/app --config platform=ios,version=0,simtype=none,devicefamily=universal,deploytype=development,target=simulator` | 579.0 ± 2.3 | 575.7 | 582.3 | 1.00 |

<details>
<summary>Benchmark command</summary>
hyperfine --prepare 'ti clean' --runs 10 'alloy compile /Users/awam/Documents/titanium-workspace/test-app/app --config platform=ios,version=0,simtype=none,devicefamily=universal,deploytype=development,target=simulator' 'alloy2 compile /Users/awam/Documents/titanium-workspace/test-app --config platform=ios,version=0,simtype=none,devicefamily=universal,deploytype=development,target=simulator'
</details>

However, there are some potential drawbacks here:

* SWC is Rust based and ships native binaries, this would be a new thing for the Alloy package so might need some testing across our supported OS's
* SWC[ has plans](https://github.com/swc-project/swc/discussions/7673) to deprecate the JS plugin method used currently and instead use [Rust based plugins](https://github.com/swc-project/swc/discussions/3540). Whilst it seems like there could be a further speed increase here, the plugin creation and distribution is tougher.
* The source maps are entirely untested and this historically has proved difficult. Given these are crucial for debugging Alloy apps we can't really compromise here, I plan to test this in the near term

If this proves useful, and there is consensus we could potentially make this an opt-in thing via a config setting to allow testing over time before making it the default.

There are some areas also that could be investigated to improve:

* If we've already copied all of the builtin files, don't run the builtins plugin
* Given we do similar work in the SDK, we could potentially replace the usage of Babel there with SWC
* Investigate allowing TypeScript usage (typechecking is not possible by SWC)

#### Testing

If you're curious in testing this you can install it using `npm i tidev/alloy#feat/use-swc -g`. Please provide any feedback you may have!


ewanharris:
@m1ga ha that was fast! Good idea on bumping the version, just pushed a commit to do that. Your numbers look inline with my tests so far, there's nothing groundbreaking in a speedup here it seems

ewanharris:
Closing this as I think it's not the best solution

PR #1366, opened 2023-10-21
chore: update Alloy defaults

m1ga:
Have some better defaults when working with Alloy:

* remove the "metal" part from the readme
* only keep one (useful) TSS class example in the default TSS comment. Most users will remove it right away. 
* make `var args` to `const args`
* add a `Window` instead of a `View` to the default View. This will run when you create a new controller and in my opinion it's more likely that you are creating a Window instead of just a View. 
  * it still needs a class otherwise it won't create a TSS file
* remove the color from the widget default. Will always mess with dark mode


ewanharris (approved):

m1ga:
Awesome, thanks for merging it 👍

PR #1368, opened 2024-01-08
chore: remove danger and jenkins related files

ewanharris:
These are no longer used

Going to repurpose this into general npm updates due to slow build times


m1ga:
Looks fine. local tests are still running but no error so far (it's a devDep anyways). Can you disable `dependabot` too? don't like all the auto PRs :smile:

m1ga (approved):
LGTM

PR #1377, opened 2024-01-15
chore: update Backbone & npm modules

m1ga:
For the next release:

* include Backbone 1.6.0 (default is still 1.4.0): https://backbonejs.org/#changelog
* update moment.js to 2.30.1: https://github.com/moment/moment/blob/develop/CHANGELOG.md
* add update info to the README


m1ga:
good point! It came out a month after I made the PR :smile: 

Tested it with an app that is using Backbone collections/models by switching to 1.6.0 in the config.json and everything still works as usual.

`npm run test` also ran without issues (`2361 specs, 0 failures`)

PR #1383, opened 2024-02-12
fix: iOS optionDialog cancel option

m1ga:
Alloy:

```xml
<Alloy>
    <Window>
        <!-- some other modules, actionbar, views and a listview -->

        <OptionDialog id="dialog" title="Delete item?" onClick="onClickDialog">
            <Options>
                <Option>Yes</Option>
                <Option cancel="true">No</Option>
            </Options>
        </OptionDialog>
    </Window>
</Alloy>
```

generated
```js
var __alloyId22 = [];
__alloyId22.push("Yes");
__alloyId19 = __alloyId22.push("No") - 1;                    // <- error
$.__views["dialog"] = Ti.UI.createOptionDialog({
  options: __alloyId22,
  cancel: __alloyId19,
  id: "dialog",
  title: "Delete item?"
});
```

with
```
[ERROR] TiExceptionHandler:
[ERROR] TiExceptionHandler: ReferenceError: __alloyId19 is not defined
[ERROR] TiExceptionHandler:     at new Controller (/alloy/controllers/machine.js:325:6199)
[ERROR] TiExceptionHandler:     at Object.exports.createController (/alloy.js:427:8)
[ERROR] TiExceptionHandler:     at Button.onClickGo (/alloy/controllers/index.js:251:21)
[ERROR] TiExceptionHandler:     at Button.value (ti:/kroll.js:1430:27)
[ERROR] TiExceptionHandler:     at Button.value (ti:/kroll.js:1477:25)
```

The `<OptionDialog>` works fine in a blank project :shrug:  

This PR will add a `var` infront of `__alloyId19`:

```js
var __alloyId22 = [];
__alloyId22.push("Yes");
var __alloyId19 = __alloyId22.push("No") - 1;                    // <- fix
$.__views["dialog"] = Ti.UI.createOptionDialog({
  options: __alloyId22,
  cancel: __alloyId19,
  id: "dialog",
  title: "Delete item?"
});
```

and the projects builds fine


hansemannn:
I am not confident to review this properly, as it's a change directly in the core compiler. Maybe @cb1kenobi or @janvennemann can help here?

Alloy/commands/compile/parsers/Alloy.Abstract.Option.js
```diff
@@ -25,7 +25,7 @@ function parse(node, state, args) {
 
 	if (attrName) {
 		if (args.createArgs[attrName]) {
-			code = `${attrVarName} = ${codePush} - 1`;
+			code = `var ${attrVarName} = ${codePush} - 1`;
 		} else {
 			code = `${attrVarName} = undefined; ${codePush}`;
```
janvennemann, line 30: I'm pretty sure this has to do with strict mode. Without the `var` this will create a new property on the global object. I guess we were not enforcing strict mode when this change was implemented six years ago. Now this throws an error in strict mode.

You should update the other variable assignment below with the same fix. Then we have all cases covered.

janvennemann (changes requested):
Looks good to me. This is probably

Alloy/commands/compile/parsers/Alloy.Abstract.Option.js
m1ga: ah, that makes sense! I was wondering why it would make a difference not having it declared with var. I've added it for the else case too

hansemannn (approved):

PR #1401, opened 2024-05-18
feat: reduce empty lines in controllers

m1ga:
before - after
Screenshot_20240518_225302

nothing major but reduces some empty lines in the generated controllers, for the looks :smile:


m1ga:
local tests:
```
Finished in 354.52 seconds
2361 specs, 0 failures
```

but I do see the
```
Warning: writing junit report failed for '', 'TEST-alloygenerate.xml'. Reasons:
  Rhino attempt: java is not defined
  PhantomJs attempt: window is not defined
  NodeJS attempt: ENOENT: no such file or directory, mkdir
```
messages too.

hansemannn (approved):

hansemannn:
Where are they generated in the first place? Would be better to not even generate them instead of manually fixing it down the road

m1ga:
This one: https://github.com/tidev/alloy/blob/master/Alloy/commands/compile/ast/controller.js#L48 is the last code that is run in Alloy which produces the newlines. I can add `.trim()` there already and the new lines are gone too. So babel is adding newlines somewhere.

Having the `.trim()` in the compile/index file was a bit of a "safety" spot in case the other methods in the AST controller produced newlines too.

m1ga:
ohhhh. I just saw that there is an option `retainLines` (https://github.com/tidev/alloy/blob/master/Alloy/commands/compile/ast/controller.js#L10) and it's true because of: https://github.com/tidev/alloy/pull/904
That will keep the new lines for debugging purposes. Sadly the ticket is not there https://jira-archive.titaniumsdk.com/TIMOB-19848 (guess it was private)

Alloy/commands/compile/compilerUtils.js
```diff
@@ -952,8 +952,9 @@ exports.loadController = function(file) {
 		U.die('Error reading controller file "' + file + '".', e);
 	}
 
+	var isProduction = (compilerConfig.alloyConfig && compilerConfig.alloyConfig.deploytype === 'production');
```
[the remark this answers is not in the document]
m1ga: it's lowercase internally: https://github.com/tidev/alloy/blob/master/Alloy/commands/compile/compilerUtils.js#L914-L919

`compilerConfig.alloyConfig?.deploytype === 'production'` will produce `955:48  error  Parsing error: Unexpected token .` during linting. Not sure how to allow that (guess an eslint package update?)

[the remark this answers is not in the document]
m1ga: found it: `'es2020': true` in eslintrc

.eslintrc.js
```diff
@@ -5,7 +5,8 @@ var ERROR = 2;
 module.exports = {
 	env: {
 		'node': true,
-		'es2017': true
+		'es2017': true,
+		'es2020': true
```
hansemannn, line 9: Not sure if this has any side effects (cc @cb1kenobi)

cb1kenobi: This shouldn't be a problem. Alloy is using the old eslint, but when we upgrade, this config will change and the whole `env` section is different.

Alloy/commands/compile/ast/controller.js
```diff
@@ -6,16 +6,20 @@ var U = require('../../../utils'),
 
 var isBaseControllerExportExpression = types.buildMatchMemberExpression('exports.baseController');
 
-const GENCODE_OPTIONS = {
+var GENCODE_OPTIONS = {
```
cb1kenobi, line 9: We should probably use the modern equivalent:

```suggestion
let GENCODE_OPTIONS = {
```

cb1kenobi (approved):

PR #1410, opened 2024-07-02
feat: download _alloy_ db

m1ga:
Run `alloy db get` to download the `_alloy_` database to `bundle.id.db`

Only tested it on Android/Linux so far. It will use `adb` to download the file. Need to test it on iOS/Mac

### DIFF
--- a/Alloy/commands/db/index.js
+++ b/Alloy/commands/db/index.js
@@ -1,0 +1,32 @@
+const {
+	exec
+} = require('child_process');
+const U = require('../../utils');
+const tiapp = require('../../tiapp');
+
+module.exports = async function(args, program) {
+
+	try {
+		if (args.length === 0) {
+			execCommand('db');
+		} else {
+			args.forEach(command => {
+				switch (command) {
+					case 'get':
+						tiapp.init();
+						console.log('Downloading _alloy_ database to: ' + tiapp.getBundleId() + '.db');
+						execCommand('adb -d shell "run-as ' + tiapp.getBundleId() + ' cat /data/data/' + tiapp.getBundleId() + '/databases/_alloy_" > ' + tiapp.getBundleId() + '.db');
+						break;
+				}
+			});
+		}
+	} catch (error) {
+		//
+	}
+};
+
+function execCommand(currentCommand) {
+	exec(currentCommand, (error, response) => {
+		return console.log(response);
+	});
+}
--- a/Alloy/tiapp.js
+++ b/Alloy/tiapp.js
@@ -40,6 +40,12 @@ tiapp.getSdkVersion = function() {
 		}
 	}
 };
+tiapp.getBundleId = function() {
+	var elems = doc.documentElement.getElementsByTagName('id');
+	if (elems && elems.length > 0) {
+		return U.XML.getNodeText(elems.item(elems.length - 1));
+	}
+};
 function getSdkSelectVersion() {
 	var homeDir = process.env[(process.platform == 'win32') ? 'USERPROFILE' : 'HOME'],
 		file = path.join(homeDir, '.titanium', 'config.json');
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-node@v1
         with:
-          node-version: '16.x'
+          node-version: '18.x'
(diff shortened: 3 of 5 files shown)


m1ga:
local tests run fine:
```
Finished in 277.454 seconds
2361 specs, 0 failures
```
online I'm getting: 
`Error:  Titanium configuration file does not exist at "/Users/runner/.titanium/config.json"`

cb1kenobi:
I have no idea why Alloy even has a database.

m1ga:
`_alloy_` is where Backbone stores all it's data by default
https://github.com/tidev/alloy/blob/2da45f850dc44473065560b552b39a6eca9af6b7/Alloy/lib/alloy/sync/sql.js#L6

Alloy/tiapp.js
```diff
@@ -39,6 +39,34 @@ tiapp.getSdkVersion = function() {
...
+
+function getSdkSelectVersion() {
+	var homeDir = process.env[(process.platform == 'win32') ? 'USERPROFILE' : 'HOME'],
+		file = path.join(homeDir, '.titanium', 'config.json');
+	if (!fs.existsSync(file)) {
+		U.die('Titanium configuration file does not exist at "' + file + '"');
+	}
+	var ticonfig = JSON.parse(fs.readFileSync(file, {encoding: 'utf8'}));
+	return ticonfig.sdk.selected;
+}
```
cb1kenobi: Dead code. Besides, there is no selected SDK anymore. It's either the `<sdk-version>` in the `tiapp.xml`, the version specified using `--sdk`, or the latest installed stable release with the last two being Titanium CLI specific, not Alloy.
```suggestion
```

Alloy/commands/db/index.js
```diff
@@ -0,0 +1,68 @@
...
+const tiapp = require('../../tiapp');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+module.exports = async function(args, program) {
+	try {
+		if (args.length === 0) {
+			U.die('Missing parameter "get"');
+		} else {
```
cb1kenobi: Early return so we don't need the `else` and indent 30-some lines of code.

Alloy/commands/db/index.js
```diff
@@ -0,0 +1,68 @@
...
+			args.forEach(command => {
```
cb1kenobi: Instead of looping to see if `'get'` is present, use `if (args.includes('get') {}`.

Alloy/commands/db/index.js
```diff
@@ -0,0 +1,68 @@
...
+				switch (command) {
+					case 'get':
+						tiapp.init();
+						var adbPath = 'adb';
+						if (os.platform() === 'darwin') {
+							// default path
+							adbPath = '~/Library/Android/sdk/platform-tools/adb';
```
cb1kenobi: Fun fact, I noticed that the `sdk` folder was at one point spelled `Sdk` and broke on case-sensitive filesystems. I have no idea if that's still the case. macOS is case-insensitive by default. I tried case-sensitive once and it broke Adobe Photoshop! I had to completely reinstall macOS again.

Alloy/commands/db/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+							// try to get android.sdkPath from ti config
+							const output = execSync('ti config android.sdkPath --json');
+							const jsonObject = JSON.parse(output);
+							if (!Object.prototype.hasOwnProperty.call(jsonObject, 'success')) {
```
cb1kenobi:
```suggestion
							if (!Object.hasOwn(jsonObject, 'success')) {
```

Alloy/commands/db/index.js
m1ga: done

Alloy/commands/db/index.js
m1ga: done

Alloy/commands/db/index.js
m1ga: good point. Added that 👍

m1ga:
oh no...I've merged a different branch into this PR :facepalm: 
Will clean that up...

m1ga:
sorry about that! Now it's clean again. Thanks for the feedback :+1:

cb1kenobi (approved):
Wow, well this PR looks completely different now! :)

PR #1420, opened 2024-09-30
fix models length issue

m1ga:
A bit difficult to test but in my case I had
```xml
<ListSection id="section" dataCollection="gyms" dataFilter="filterGyms" dataTransform="transformGyms" dataFunction="setListData">
```

which produced: 

```js
var models = filterGyms(__alloyId1172);
var len = models.length;
```

and models was `null` for me which caused an error "length of undefined". 

Instead of crashing it should just use `len = 0` for this.

merge https://github.com/tidev/alloy/pull/1421 first. There I'm fixing build issues and some other Alloy code


m1ga:
@cb1kenobi any idea why the tests are failing? They work fine locally
```
Error:  Titanium configuration file does not exist at "/Users/runner/.titanium/config.json"
```
not sure how I can fix that.

cb1kenobi:
That's not a CLI error, it's an Alloy error: https://github.com/tidev/alloy/blob/be399c89e6e1123a3aa1428996558597a5060cf5/Alloy/tiapp.js#L47.

Alloy is reading the config to get the selected sdk, but that's obsolete in CLI v7. Alloy should get the SDK version from the `cli` object passed into the hook. Alloy should become apart of the CLI/SDK, not some external thing.

m1ga:
ohhhh ok, I didn't check for the error message in the project :facepalm: Let me fix that! Thanks for the hint

m1ga:
merge https://github.com/tidev/alloy/pull/1421 first. There I'm fixing build issues and some other Alloy code

cb1kenobi (approved):
LGTM. Approving, but waiting for PR #1421 to land.

cb1kenobi (approved):

PR #1421, opened 2024-10-01
remove old code, fix sdk version, update action

m1ga:
* update github action
* remove <= Ti 10 if cases
* remove `ti config` selected.sdk check (not used anymore)
* add ti sdk version to the test tiapp.xml file
* use 12.5.0 in the github action so it will work with the test


.github/workflows/build.yml
```diff
@@ -11,12 +11,12 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v4
         with:
-          node-version: '16.x'
+          node-version: '18.x'
```
cb1kenobi, line 19: Please use Node 22. We are dropping Node 18 in the next major CLI release.

.github/workflows/publish.yml
```diff
@@ -8,14 +8,14 @@ jobs:
...
+      uses: actions/checkout@v4
       with:
         fetch-depth: 0
 
     - name: Setup node
-      uses: actions/setup-node@v2
+      uses: actions/setup-node@v4
       with:
-        node-version: '16'
+        node-version: '18'
```
cb1kenobi, line 18: Same, let's use Node 22 please.

test/projects/HarnessTemplate/tiapp.xml
```diff
@@ -69,4 +69,5 @@
         <module platform="android">ti.map</module>
         <module platform="iphone">ti.map</module>
     </modules>
+    <sdk-version>12.5.0.GA</sdk-version>
```
cb1kenobi, line 72: If you omit `<sdk-version>`, it'll use the latest installed SDK. This is handy so we don't have to remember to bump this.

cb1kenobi (changes requested):
I love this PR! Needs a couple of small changes.

What's up with the removal of `Ti.UI.iPhone.NavigationGroup.js`? Is this a breaking change?

test/projects/HarnessTemplate/tiapp.xml
m1ga: The test failed without a version and that's why I've put 12.5.0 in there and made sure 12.5.0 is installed in the build workflow

cb1kenobi: The Titanium CLI doesn't need a `<sdk-version>`. I built an alloy app without a `<sdk-version>` and it also worked.

I tried running the tests on Windows and they blow up saying `'alloy' is not recognized as an internal or external command`. `alloy` is in the PATH. I haven't dug further into it.

Maybe leave this as `latest`?

m1ga: This is the error when I use latest and remove the `<sdk-version>` again:

Bildschirmfoto_20241007_104735

```
Error:  "undefined" was found in the "sdk-version" field of your tiapp.xml.
Error:  If you are building with the legacy titanium.py script and are specifying 
Error:  an SDK version as a CLI argument that is different than the one in your 
Error:  tiapp.xml, please change the version in your tiapp.xml file.
```

cb1kenobi: I'll test on Linux when I get a sec.

cb1kenobi: Ok, looks like you backed out the changes, so we're good.

cb1kenobi:
The tests fail on WSL for the same reason the CI checks fail below. For example:

```
alloy compile widgets/wpath/ [ios] compiles without critical error
  Expected { code : 1, killed : false, signal : null, cmd : 'alloy compile /home/chris/projects/titanium/alloy/test/projects/Harness --config platform=ios' } to be falsy.
```

Not sure what's causing that. Isn't the Android and iOS stuff mocked?

cb1kenobi (approved):

m1ga:
@cb1kenobi did you had some time to look at this? I can put in the fixed <sdk-version> again so the test will succeed

m1ga:
@cb1kenobi I've added the `sdk-version` back to the test tiapp.xml so everything is green and we can get this merged + #1420 (and #1410 in the future as it is very hand to get the database if you work with models :) ).

cb1kenobi (approved):

PR #1423, opened 2025-05-01
fix: backbone 1.6.0 fixes

m1ga:
when switching from Backbone 1.5.0 to 1.6.0 in the last PR I forgot to update these two files in order to use underscore functions.

I most Backbone stuff was working but functions like `findWhere` didn't work.

After this fix everything is working again


m1ga:
@hansemannn you might want to check this too. It's in the readme (https://github.com/tidev/alloy?tab=readme-ov-file#backbone) but I forgot to change it when I've moved from 1.5.0 to 1.6.0 in the last PR.

After that we should be fine for a release :+1:

hansemannn (approved):

PR #1427, opened 2025-07-19
On publish post release notification to Slack

cb1kenobi:


.github/workflows/publish.yml
```diff
@@ -22,10 +20,47 @@ jobs:
...
       if: steps.node-cache.outputs.cache-hit != 'true'
 
+    - name: Get current version
+      id: get-current-version
+      run: echo "version=$(node -p "require('./package.json').version")" >> $GITHUB_OUTPUT
+
     - name: Publish to npm
       env:
         GH_TOKEN: ${{ github.token }}
-        GITHUB_TOKEN: ${{ github.token }}
```
cb1kenobi, line 28: Are `GITHUB_TOKEN` and `NPM_TOKEN` are used? titanium-editor-commons doesn't have them: https://github.com/tidev/titanium-editor-commons/blob/master/.github/workflows/publish.yml#L32-L33.

PR #1441, opened 2026-02-26
feat: add missing parser files

m1ga:
Adding default parser files for some UI elements.

The autocompletion parsers use these files to generate the `completions-v3.json`. Currently you can't type `<Acti` to get the autocompletion for `<ActivityIndicator>`. This PR will add those files and when you clear your `.titanium/completions/titanium/13.1.1.GA/` files and let Pulsar rebuild those it will find the missing elements.


cb1kenobi (approved):
This is such a strange pattern for defining a callback. I can't wait to rewrite this.

PR #1442, opened 2026-03-13
Remove more windows/mobileweb parts

m1ga:
* remove more `windows` and `mobileweb` parts
* empty `plugin.py` (can't remove it as node-appc requires it to be there, but it is not used)

`npm run test` output:
> Finished in 238.308 seconds
> 2778 specs, 0 failures

**Note:** 
I've kept `platforms/mobileweb/index.js` and `platforms/mobileweb/index.js` as it will break kitchensink and any other app that still has e.g. `platform=mobileweb` in it. As soon as we have all examples in the tidev repo clean I would say that we remove those two files.


cb1kenobi (approved):
Awesome!

PR #1444, opened 2026-04-12
feat: allow comma separated class selectors

m1ga:
This PR will allow the user to have selectors like `'.btn, .btn2'` (like in HTML)
Need to do some more testing and see if there are no other cases where a comma is allowed in the selector. 
```xml
<Alloy>
  <Window layout="vertical">
    <Button class="btn">Button 1</Button>
    <Button class="btn2">Button 2</Button>
  </Window>
</Alloy>
```
```
'.btn, .btn2': {
  backgroundColor: 'red'
}

'.btn2': {
  color: 'green'
}
```

Screenshot_20260412-143206

The later style will override the parts of the first so 
```
'.btn, .btn2': {
  backgroundColor: 'red',
  color: 'blue'
}

'.btn2': {
  color: 'green'
}
```

will only make the first button text color blue.

All tests pass: 2778 specs, 0 failures
two bigger Alloy apps still look fine, so nothing broke in existing styles.

### DIFF
--- a/Alloy/commands/compile/styler.js
+++ b/Alloy/commands/compile/styler.js
@@ -195,9 +195,13 @@ exports.sortStyles = function(style, opts) {
 	var sortedStyles = [];
 	opts = opts || {};
 
+	function splitSelectors(key) {
+		return key.split(/,(?![^[]]*\])/).map(function(k) { return k.trim(); }).filter(Boolean);
+	}
+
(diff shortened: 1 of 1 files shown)


ewanharris (approved):
Changes look good to me with the clone addition, could we maybe add some tests to ensure that behaviour can be defined?

m1ga:
I've pushed some of the test files I was using before and let claude add some more test cases :smile:

PR #1445, opened 2026-05-10
code updates and fixes

m1ga:
**Fixes:**

* remove all **path.existsSync** (see comment in current Alloy/alloy.js: patch to remove the warning in node >=0.8)
* **Alloy/builtins/social.js**: ++s instead of ++i in percentEncode — causes guaranteed app hang when called with Array arg
* **Alloy/commands/compile/Orphanage.js**: Inner loop uses outer variable i instead of j — wildcard file exclusion matching is completely broken
* **Alloy/lib/alloy/sync/sql.js**: DB cursor leak on exception — rs.close() never called if loop body throws

**Performance improvments:**
* **Cached readdirSync in per-element hot loop**
**compilerUtils.js** — The parsers directory was listed on every XML element (hundreds of times per compile). Now cached at module init in parserFileNames.
Before: fs.readdirSync(parsersDir) inside generateNode() (every XML element)  
After: parserFileNames populated once at load via IIFE, used by _.includes()

* **Replaced eval() with property chain walk**
**template/lib/alloy.js:228** — eval(ns) prevented V8 optimization for every Alloy.UI.create() call. Replaced with resolveNS() that walks the global object property chain via a simple for loop.
Before: `return eval(ns)['create' + baseName](style);  `
After: `return resolveNS(ns)['create' + baseName](style);`

* **Cached compiled template functions**
**index.js** — Three locations were reading template files from disk and calling _.template() (compile step) on every component iteration. Now cached in templateCache via getCompiledTemplate().
Files cached:
   - **wpath.js** — was read+compiled at lines 608 and 1013 per widget component
   - **component.js** — was read+compiled at line 896 per view/controller
Before: _.template(fs.readFileSync(...)) repeated per iteration  
After: getCompiledTemplate(name) reads+compiles once, returns cached function

* **Removed the redundant babylon.parse sanity check in generateSourceMap** (lines 217-231): Generated code is parsed into an AST purely as validation, then babel.transformFromAstSync parses it again internally. The first parse is wasted CPU. It parsed the code into an AST purely to catch syntax errors, then discarded it without ever using it. The TODO comment even asked "Remove? This is a sanity check...", confirming it was dead work.

All tests are still working: 2778 specs, 0 failures
and I've tested two apps with and without liveview and they look and behave the same.

On a very large Alloy project `alloy compile --config platform=android` went from 4.9s to 4.6s

### DIFF
--- a/Alloy/alloy.js
+++ b/Alloy/alloy.js
@@ -14,8 +14,7 @@ var program = require('commander'),
 	fs = require('fs'),
 	CONST = require('./common/constants');
 
-// patch to remove the warning in node >=0.8
-path.existsSync = fs.existsSync || path.existsSync;
+
 
 // avoid io issues on Windows in nodejs 0.10.X: https://github.com/joyent/node/issues/3584
 if (process.env.ALLOY_TESTS && /^win/i.test(os.platform())) {
@@ -144,7 +143,7 @@ function getCommands() {
 	try {
 		var commandsPath = path.join(__dirname, 'commands');
 		return _.filter(fs.readdirSync(commandsPath), function(file) {
-			return path.existsSync(path.join(commandsPath, file, 'index.js'));
+			return fs.existsSync(path.join(commandsPath, file, 'index.js'));
 		});
 	} catch (e) {
 		U.die('Error getting command list', e);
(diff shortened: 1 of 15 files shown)


cb1kenobi (approved):
Excellent work!

PR #1446, opened 2026-05-12
fix: attach ES6 named exports to controller exports

hansemannn:
## Summary

Controllers written with ES6 export syntax (`export function show() {}`, `export const show = () => {}`, `export class X {}`, `export { show }`) silently dropped their exports — only the CommonJS form `exports.show = () => {}` worked. The controller methods existed in the generated file but were never attached to the controller instance, so calling `$.show()` from outside would be `undefined`.

Tested with our app as well!

## Root cause

`Alloy/commands/compile/ast/controller.js` handled `ExportNamedDeclaration` by stripping the entire node out of the controller body and dumping it verbatim into `moduleCodes`. That string is interpolated into the generated component template at **file-module** scope (`Alloy/template/component.js:5`), *above* the `Controller` function.

Inside `Controller()` there is a *local* `var exports = {}` (line 28) that the developer's controller body is supposed to populate; `_.extend($, exports)` (line 55) is what surfaces those names on the controller instance. Because the ES6 exports were hoisted out of the function, they never wrote to this local object and were lost — `module.exports = Controller` at the bottom of the template overrides anything assigned at module scope anyway.

CommonJS `exports.show = ...` worked because it's a plain `AssignmentExpression`, not an `ExportNamedDeclaration`, so no AST handler touched it and it stayed in the controller body.

## Fix

Rewrote the `ExportNamedDeclaration` handler to transform local exports into a declaration + `exports.NAME = NAME` assignment pair, keeping both in the controller body:

- `export function show() {}` → `function show() {} exports.show = show;`
- `export const show = () => {}` → `const show = () => {}; exports.show = show;`
- `export const a = 1, b = 2;` → two assignments
- `export class Foo {}` → `class Foo {} exports.Foo = Foo;`
- `export { show }` → `exports.show = show;` (declaration left in place)
- `export { show as displayShow }` → `exports.displayShow = show;`
- `export { foo } from './bar'` (has a `source`) → still hoisted to module scope as before — it's a true module-level construct
- CommonJS `exports.show = ...` → unchanged

The obsolete second AST traversal that yanked function/variable declarations out of the controller body has been removed.

## Test plan

- [x] CommonJS `exports.show = () => {}` still inlined unchanged (regression check)
- [x] `export function show() {}` produces `function show() {} exports.show = show;`
- [x] `export const show = () => {}` produces `const show = () => {}; exports.show = show;`
- [x] `export const a = 1, b = 2;` produces both `exports.a` and `exports.b` assignments
- [x] `export class Foo {}` produces class declaration + `exports.Foo = Foo`
- [x] `export { show }; function show() {}` keeps the function and adds `exports.show = show`
- [x] `export { show as displayShow }` renames correctly
- [x] `export { foo } from './bar'` stays hoisted to module scope
- [x] Combined `import` + `export` keeps `import` at module scope and rewrites the export
- [x] `exports.baseController = "X"` detection still works
- [x] `npx eslint Alloy/commands/compile/ast/controller.js` clean
- [x] Run an Alloy app that uses `export function` syntax and confirm the method is callable on the controller instance

### DIFF
--- a/Alloy/commands/compile/ast/controller.js
+++ b/Alloy/commands/compile/ast/controller.js
@@ -13,11 +13,20 @@ let GENCODE_OPTIONS = {
 exports.processController = function(code, file, isProduction = false) {
 	var baseController = '',
 		moduleCodes = '',
-		newCode = '',
-		exportSpecifiers = [];
+		newCode = '';
 
 	if (isProduction) {
 		GENCODE_OPTIONS.retainLines = false;
+	}
+
+	function buildExportAssignment(exportedName, localName) {
+		return types.expressionStatement(
+			types.assignmentExpression(
+				'=',
+				types.memberExpression(types.identifier('exports'), types.identifier(exportedName)),
+				types.identifier(localName)
+			)
+		);
 	}
 
 	try {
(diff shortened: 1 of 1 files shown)


hansemannn:
@cb1kenobi With https://github.com/tidev/titanium-vite, this will soon not be necessary again, but as it will be opt-in in the beginning, this solution lays the ground work to have proper ES6-based exports already, so future tooling upgrades will be easier to adopt :-)

cb1kenobi (approved):
I love this, awesome!

PR #1447, opened 2026-07-02
code improvements

m1ga:
**1. `BaseController.removeListener()` — Memory leak via sparse array**

`removeListener` used `delete self.__events[index]` inside a `forEach`, which creates holes in the array. The array never shrinks, so `__events.length` grows monotonically over the app lifecycle, degrading `getListener()` and any iteration over tracked events.

**Fix:** Reverse `for` loop with `splice(i, 1)` — removes the element properly, collapses the array, and avoids index-shifting issues during iteration.

---

**2. `BaseController.destroy()` — Tracked event listeners not cleaned up**

The `destroy()` method generated at compile time only handled model/collection binding cleanup. Event listeners registered via `$.addListener()` (or XML markup) persisted in `__events` even after the controller was destroyed, causing memory leaks.

**Fix:** Added `cleanup()` to BaseController that removes all tracked event listeners and resets `__events`. The compiler now prepends `this.cleanup()` to the generated `destroy()` body, so `$.destroy()` automatically cleans up all listeners with no developer action required.

---

**3. `alloy.js deepExtend()` — Slow runtime style merging**

The custom deep-merge used for every `$.createStyle()` / `$.addClass()` call had several inefficiencies:
- Recursion via `exports.deepExtend` (property lookup on every recursive call)
- `for...in` without `hasOwnProperty` guard (risks prototype pollution and extra iterations)
- Unused dead code path splitting strings into char arrays
- `_.isObject()`/`_.isFunction()` function calls where inline `typeof` would suffice

**Fix:** Named function expression for direct internal recursion, `hasOwnProperty` guard on `for...in`, removed dead code, and replaced `_.isObject` with `typeof copy === 'object'` for the hot path.

---

All tests work fine: `Finished in 469.757 seconds, 2778 specs, 0 failures`


Alloy/commands/compile/index.js
```diff
@@ -881,8 +881,10 @@ function parseAlloyComponent(view, dir, manifest, noView, fileRestriction) {
 			handlerVar + ');';
 	});
 
-	// add destroy() function to view for cleaning up bindings
-	template.viewCode += 'exports.destroy = function () {' + CU.destroyCode + '};';
+	// add destroy() function to view for cleaning up bindings and tracked
+	// event listeners. Use `$` rather than `this` so destroy() still works
+	// when invoked detached, e.g. `$.win.addEventListener('close', $.destroy)`.
+	template.viewCode += 'exports.destroy = function () {$.removeListener();' + CU.destroyCode + '};';
```
ewanharris, line 887: Is there a risk of the behaviour change of calling removeListener impacting peoples apps?

Alloy/template/lib/alloy.js
```diff
@@ -602,64 +603,63 @@ if (OS_ANDROID) {
...
+ * and are not recursed into. Note that this is a heuristic: proxies without an
+ * own `apiName` are merged like plain objects.
+ *
+ * @param {Boolean} deep  If true, recurse into plain objects and arrays.
+ * @param {Object}  target  The object to receive merged properties.
+ * @param {...Object} sources  One or more source objects.
+ * @return {Object}  The target object, mutated in place.
  */
-exports.deepExtend = function() {
+exports.deepExtend = function deepExtend() {
```
ewanharris, line 620: In the description it calls out a `hasOwnProperty` check? Maybe add the check alongside the existing `name` check?

Alloy/commands/compile/index.js
m1ga: I was only finding this (scroll down to the warning box):

https://titaniumsdk.com/guide/Alloy_Framework/Alloy_Guide/Alloy_Models/Alloy_Data_Binding.html#collection-view-binding

that says: `When using Alloy's data binding in a view-controller, you MUST call the $.destroy() function when closing a controller to prevent potential memory leaks`. So in theory `$.destroy()` should never been called before so we are in a state where the controller is closing anyways.

Alloy/template/lib/alloy.js
m1ga: added the name check again.

Some test code:

```js
var results = [];

function check(ok, msg) {
	results.push((ok ? 'PASS ' : 'FAIL ') + msg);
	Ti.API.info((ok ? 'PASS ' : 'FAIL ') + msg);
}

var r = Alloy.deepExtend(true, {}, {
	a: 1,
	n: {
		x: 1
	}
}, {
	n: {
		y: 2
	},
	b: [1, 2]
});
check(JSON.stringify(r) === '{"a":1,"n":{"x":1,"y":2},"b":[1,2]}', 'plain deep merge');

Alloy.deepExtend(true, {}, JSON.parse('{"__proto__":{"polluted":1}}'));
check(({}).polluted === undefined, 'JSON __proto__ not merged');

var proto = {
	inherited: 1
};
var o = Object.create(proto);
o.own = 2;
r = Alloy.deepExtend(true, {}, o);
check(r.own === 2 && !('inherited' in r), 'inherited props skipped');

Object.prototype.zzz = 'bad';
r = Alloy.deepExtend(true, {}, {
	k: 1
});
delete Object.prototype.zzz;
check(r.k === 1 && !Object.prototype.hasOwnProperty.call(r, 'zzz'), 'polluted Object.prototype not leaked');

var view = Ti.UI.createView({
	backgroundColor: 'red',
	width: 10,
	height: 10
});
r = Alloy.deepExtend(true, {}, {
	child: view
});
check(r.child === view, 'Ti proxy kept by reference');

var d = new Date(0);
r = Alloy.deepExtend(true, {}, {
	d: d
});
check(r.d !== d && r.d.valueOf() === 0, 'Date cloned');

var lbl = Ti.UI.createLabel(Alloy.deepExtend(true, {}, {
	color: '#000',
	font: {
		fontSize: 14
	}
}, {
	font: {
		fontWeight: 'bold'
	},
	top: 20
}));
check(lbl.font.fontSize === 14 && lbl.font.fontWeight === 'bold' && lbl.top === 20, 'createLabel with merged args');

const win = Ti.UI.createWindow({layout:"vertical"});

results.forEach(function(line) {
	win.add(Ti.UI.createLabel({
		text: line,
		color: line.indexOf('PASS') === 0 ? '#080' : '#c00',
		top: 8,
		left: 12,
		width: Ti.UI.FILL,
		font: {
			fontSize: 14,
			fontFamily: 'monospace'
		}
	}));
});
win.open();
```

